PR #2, opened 2025-10-16
Add SplitPromptOnDashTransform to handle script-based multiturn prompts

richl9:
This transform is useful for script-based execution flows, allowing a single prompt script to be split into multiple conversational turns for step-by-step reasoning, leveraging the model’s strength with structured instructions. We are using script-based prompt a lot in our workflow. e.g.

example.script
```
Analyze A and do step one.

---
Now based on the result of your analysis, perform step two.

---
Now generate a json result following below format:
....

```

### DIFF
--- a/wayflowcore/src/wayflowcore/transforms.py
+++ b/wayflowcore/src/wayflowcore/transforms.py
@@ -5,7 +5,7 @@
 # 2.0 (LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0), at your option.
 
 from dataclasses import dataclass
-from typing import TYPE_CHECKING, Any, Callable, List
+from typing import TYPE_CHECKING, Any, Callable, List, Optional
 
 from wayflowcore._utils.async_helpers import is_coroutine_function, run_sync_in_thread
 from wayflowcore.serialization.serializer import SerializableCallable, SerializableObject
@@ -113,3 +113,27 @@ class AppendTrailingSystemMessageToUserMessageTransform(MessageTransform, Serial
 
         penultimate_message.contents.extend(last_message.contents)
         return messages[:-2] + [penultimate_message]
+
+
+class SplitPromptOnMarkerMessageTransform(MessageTransform, SerializableObject):
+    """
+    Split prompts on a marker into multiple messages with the same role. Only apply to the messages without tool_requests and tool_result.
+
+    This transform is useful for script-based execution flows, where a single prompt script can be converted into multiple conversation turns for step-by-step reasoning.
+    """
+
+    def __init__(self, marker: Optional[str] = None):
+        self.marker = marker if marker is not None else "\n---"
+
+    def __call__(self, messages: list["Message"]) -> list["Message"]:
+        new_messages = []
+
+        for msg in messages:
+            if msg.tool_requests is None and msg.tool_result is None and self.marker in msg.content:
+                for part in (p.strip() for p in msg.content.split(self.marker) if p.strip()):
+                    new_msg = msg.copy(content=part)
+                    new_messages.append(new_msg)
+            else:
+                new_messages.append(msg)
+
+        return new_messages
--- a/wayflowcore/tests/test_transforms.py
+++ b/wayflowcore/tests/test_transforms.py
@@ -14,6 +14,7 @@ from wayflowcore.tools import ToolRequest, ToolResult
 from wayflowcore.transforms import (
     CoalesceSystemMessagesTransform,
     RemoveEmptyNonUserMessageTransform,
+    SplitPromptOnMarkerMessageTransform,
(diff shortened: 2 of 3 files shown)


dhilloulinoracle:
Hi @richl9 ,
Thanks a lot for your contribution!

Would you mind describing in your PR description what is being added, and for what kind of use-cases it can be useful?

dhilloulinoracle:
@cesarebernardis , @paul-cayet , can you please review?

wayflowcore/src/wayflowcore/transforms.py
```diff
@@ -113,3 +113,26 @@ def __call__(self, messages: List["Message"]) -> List["Message"]:
...
+class SplitPromptOnDashTransform(MessageTransform, SerializableObject):
+    """
+    Split prompts on a marker consisting of a **newline followed by '---'** into multiple messages
+    with the same role. Only apply to the messes without tool_requests and tool_result.
+
+    This transform is useful for script-based execution flows, where a single prompt script can be converted into multiple conversation turns for step-by-step reasoning.
+    """
+
+    def __call__(self, messages: List["Message"]) -> List["Message"]:
+        marker = "\n---"
```
cesarebernardis: Could you make the `marker` an attribute of this class, so that user can customize it when they create an instance?

I think you could also change the name of the class to something a bit more generic at that point, like `SplitPromptOnMarkerMessageTransform`

wayflowcore/src/wayflowcore/transforms.py
```diff
@@ -113,3 +113,26 @@ def __call__(self, messages: List["Message"]) -> List["Message"]:
 
         penultimate_message.contents.extend(last_message.contents)
         return messages[:-2] + [penultimate_message]
+
+
+class SplitPromptOnDashTransform(MessageTransform, SerializableObject):
+    """
+    Split prompts on a marker consisting of a **newline followed by '---'** into multiple messages
+    with the same role. Only apply to the messes without tool_requests and tool_result.
```
cesarebernardis: Typo `messes` -> `messages` I think

wayflowcore/src/wayflowcore/transforms.py
```diff
@@ -113,3 +113,26 @@ def __call__(self, messages: List["Message"]) -> List["Message"]:
...
+    This transform is useful for script-based execution flows, where a single prompt script can be converted into multiple conversation turns for step-by-step reasoning.
+    """
+
+    def __call__(self, messages: List["Message"]) -> List["Message"]:
+        marker = "\n---"
+        new_messages = []
+        for msg in messages:
+            if msg.tool_requests is None and msg.tool_result is None and marker in msg.content:
+                parts = [p.strip() for p in msg.content.split(marker) if p.strip()]
+                for part in parts:
```
cesarebernardis: I think you could merge the previous for loop that creates the list and this one. Something like:

```
for part in filter(lambda p: bool(p), (p.strip() for p in msg.content.split(marker))):
```

wayflowcore/src/wayflowcore/transforms.py
```diff
@@ -113,3 +113,26 @@ def __call__(self, messages: List["Message"]) -> List["Message"]:
...
+                    new_msg = msg.copy()
```
cesarebernardis: You can pass the content directly to the copy and it will create a new instance with the new content value, without the need to set it afterwards. Something like
```
new_msg = msg.copy(content=part)
```

cesarebernardis: After applying this change I think you could also remove the property setter on the `content` attribute you added to the `Message` class

wayflowcore/tests/test_transforms.py
```diff
@@ -175,3 +176,29 @@ def test_python_merge_tool_request(messages, expected_messages):
...
+                Message(message_type=MessageType.USER, content="B"),
+                Message(message_type=MessageType.USER, content="C"),
+            ],
+        ),
+    ],
+)
+def test_split_prompt_on_dash(messages, expected_messages):
+    transform = SplitPromptOnDashTransform()
+    transformed_messages = transform(messages)
+    assert_messages_are_correct(transformed_messages, expected_messages)
```
cesarebernardis, line 204: Since `SplitPromptOnDashTransform` is a `SerializableObject` you should add it also to the serialization test. In other words, you should add it to the `ALL_SERIALIZABLE_CLASSES` set in the file `wayflowcore/tests/serialization/test_serializableobject.py`

richl9:
@cesarebernardis Thanks for the review. I have addressed all the comments.

cesarebernardis:
> @cesarebernardis Thanks for the review. I have addressed all the comments.

LGTM, Thanks for the contribution!

PR #16, opened 2026-01-09
Python 3.14 support

cesarebernardis:
Adding support for python 3.14 in WayFlow

### DIFF
--- a/wayflowcore/tests/testhelpers/test_server_port_retrieval.py
+++ b/wayflowcore/tests/testhelpers/test_server_port_retrieval.py
@@ -1,0 +1,26 @@
+# Copyright © 2026 Oracle and/or its affiliates.
+#
+# This software is under the Apache License 2.0
+# (LICENSE-APACHE or http://www.apache.org/licenses/LICENSE-2.0) or Universal Permissive License
+# (UPL) 1.0 (LICENSE-UPL or https://oss.oracle.com/licenses/upl), at your option.
+
+import uuid
+from concurrent.futures import ThreadPoolExecutor
+
+import pytest
+
+from ..utils import get_available_port
+
+
+def test_get_available_port_returns_unique_ports_in_parallel(
+    monkeypatch, tmp_path_factory: pytest.TempPathFactory
+):
+    monkeypatch.setenv("WAYFLOW_TEST_PORT_BASE", "7000")
+    monkeypatch.setenv("WAYFLOW_TEST_PORT_SPAN", "200")
+
+    temp_dir = tmp_path_factory.mktemp(f"port-retrieval-{uuid.uuid4().hex}")
+    request_count = 30
+    with ThreadPoolExecutor(max_workers=request_count) as executor:
+        futures = [executor.submit(get_available_port, str(temp_dir)) for _ in range(request_count)]
+    ports = {future.result() for future in futures}
+    assert len(ports) == request_count


nasseur:
Internal regression failed: Build ID #83

nasseur:
Internal regression failed: Build ID #88

nasseur:
Internal regression failed: Build ID #196

nasseur:
Internal regression failed: Build ID #198

nasseur:
Internal regression failed: Build ID #356

wayflowcore/tests/a2a/conftest.py
```diff
@@ -89,11 +88,12 @@ def _start_a2a_server(
 
 
 @pytest.fixture(scope="session", name="a2a_server")
-def a2a_server_fixture(request):
+@pytest.mark.filterwarnings("ignore:unclosed:ResourceWarning")
```
paul-cayet: Do we have an idea of where this is coming from? Please add comment if possible

wayflowcore/tests/test_docs_examples.py
```diff
@@ -31,8 +32,8 @@
 sys.path.insert(0, os.path.abspath(UTILS_DIR))
 
 
-@pytest.fixture()
-def mock_server():
+@pytest.fixture(scope="session")
```
paul-cayet, line 35: Please add a small comment saying what session scope is (does it mean the entire test suite or this file, or more?)

wayflowcore/tests/test_docs_examples.py
```diff
@@ -90,20 +91,22 @@ async def protected_endpoint(request: Request):
...
-            if r.status_code in {400, 200, 401}:
-                break
-        except Exception:
-            time.sleep(0.2)
-    else:
-        raise RuntimeError("Mock server did not start in time")
+    with httpx.Client(timeout=0.2) as client:
+        for _ in range(50):  # up to 5 seconds
+            try:
+                response = client.get(url, headers=headers)
```
paul-cayet, line 97: This is not the same code right? Don't we need to add a timeout on this client.get as well or is it inherited from the httpx.Client object instance? Please make sure of that to avoid hanging

wayflowcore/tests/testhelpers/testhelpers.py
```diff
@@ -281,9 +281,8 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
                         total_time_of_failed_runs += time.perf_counter() - start_time
                         signal.alarm(0)
                         time.sleep(wait_between_tries)
-                    finally:
-                        if time.time() - loop_start_time > FLAKY_TEST_MAX_EXECUTION_TIME_PER_TEST:
-                            break
+                    if time.time() - loop_start_time > FLAKY_TEST_MAX_EXECUTION_TIME_PER_TEST:
```
paul-cayet, line 284: nit: Why doing this?

wayflowcore/tests/utils.py
```diff
@@ -141,9 +141,78 @@ def _check_server_is_up(
...
+            return True
+        except OSError:
+            return False
+
+    # Acquire an interprocess lock, POSIX only. On non-POSIX, degrade to probing loop.
+    if os.name == "posix":
+        import fcntl
+
+        with open(lock_path, "a+") as lockf:
+            fcntl.flock(lockf.fileno(), fcntl.LOCK_EX)
```
paul-cayet, line 184: Please add a short comment explaining what this line is doing (e.g., I have no idea what is this .LOCK_EX)

wayflowcore/tests/a2a/conftest.py
jschweiz: from the a2a server implementation, because they use threads in the backend

wayflowcore/tests/a2a/test_a2aserver.py
```diff
@@ -9,17 +9,17 @@
 from pathlib import Path
 from typing import Optional
 
+import httpx
 import pytest
-import requests
 from fasta2a.schema import FileWithBytes
-from requests import Response
```
jschweiz, line 15: thanks

wayflowcore/tests/utils.py
```diff
@@ -141,9 +141,78 @@ def _check_server_is_up(
...
+        WAYFLOW_TEST_PORT_BASE  base port (int), default is 10000
+        WAYFLOW_TEST_PORT_SPAN  size of the port window starting at BASE (int), default is 500
+    - Within the critical section, probe ports by attempting a bind; skip ports already in use.
+
+    The parameter tmp_path is where the file used for the lock is located. It should be consistent
+    across threads in order to ensure that they acquire the lock on the same file, so we should
+    always call this method passing the value of the `session_tmp_path` fixture.
+    """
+
+    base = int(os.environ.get("WAYFLOW_TEST_PORT_BASE", 10000))
```
paul-cayet, line 162: nit: Can you rename to `base_port`?

wayflowcore/tests/integration/test_descriptors.py
```diff
@@ -110,13 +111,15 @@ def deserialization_context_with_tool():
...
     DatastoreListStep.__name__: {"limit": 1},
+    InputMessageStep.__name__: {"message_template": "Hello"},
     OutputMessageStep.__name__: {"message_type": MessageType.AGENT},
     RegexExtractionStep.__name__: {"regex_pattern": RegexPattern(pattern=".*")},
     ParallelFlowExecutionStep.__name__: {"flows": [get_single_step_flow()]},
 }
 
 # Add step class names in this list if they are not meant to be actually tested
 step_cls_skiplist = [
+    "Step",
```
jschweiz, line 122: not sure why this is needed? But doesn't seem to harm anything

wayflowcore/tests/utils.py
```diff
@@ -141,9 +141,78 @@ def _check_server_is_up(
...
+    span = int(os.environ.get("WAYFLOW_TEST_PORT_SPAN", 500))
```
paul-cayet, line 163: nit: can you rename for `port_span`?

wayflowcore/tests/run_tests.sh
```diff
@@ -15,7 +15,7 @@ if [ "$1" = "--parallel" ]; then
     # Run tests in parallel
     # disable a2a server tests (issue #40)  # tests/a2a
     pytest tests/test_docstring.py tests/datastores tests/agentserver tests/transforms/test_summarization_transforms.py
-    pytest -n auto $REPO_ROOT/tests --dist loadscope --ignore=tests/datastores/ --ignore=tests/test_docstring.py --ignore=tests/agentserver/ --ignore=tests/transforms/test_summarization_transforms.py --ignore=tests/a2a
+    pytest -n auto --maxprocesses=4 $REPO_ROOT/tests --dist loadscope --ignore=tests/datastores/ --ignore=tests/test_docstring.py --ignore=tests/agentserver/ --ignore=tests/transforms/test_summarization_transforms.py --ignore=tests/a2a
```
jschweiz: this is not needed if we don't run the CI in parallel

wayflowcore/src/wayflowcore/tools/toolhelpers.py
```diff
@@ -77,7 +78,7 @@ def _get_partial_schema_from_annotation(arg_type: Type[Any]) -> JsonSchemaParam:
         }
 
     # Handle Union[X, Y, Z]
-    if origin is Union:
+    if origin in (UnionType, Union):
```
paul-cayet, line 81: Can you add a tests to make sure that this works on this pipe-type union annotation

e.g.
```python
value: str | bool
```
Thanks

wayflowcore/tests/test_docs_examples.py
jschweiz: this is basic pytest, and would depend if you run with multiple workers or not, so no need to clutter

wayflowcore/tests/test_docs_examples.py
```diff
@@ -90,20 +91,22 @@ async def protected_endpoint(request: Request):
...
+                time.sleep(0.2)
+        else:
+            raise RuntimeError("Mock server did not start in time")
+
     try:
         yield f"http://{host}:{port}/protected"
     finally:
         server.should_exit = True
-        thread.join(timeout=5)
+        thread.join(timeout=3.0)
```
jschweiz: is that enough? The CI can be pretty slow, so better be safe and slow than sorry

wayflowcore/tests/test_docs_examples.py
```diff
@@ -133,8 +136,14 @@ def _update_globals(varnames_to_update: List[str]) -> Tuple[Any, ...]:
     return _update_globals
 
 
+# We retry code example tests as some of them rely on LLMs, but we keep the retry number low.
+# We should not increase the max attempts beyond 2 because flakiness of docs should be minimal.
+# If you are considering raising this number, please make your documentation example more robust instead.
+@retry_test(max_attempts=2)
```
jschweiz, line 142: do we even want the docs to be flaky? Sounds like a very bad experience if the users can get their "get started" code to crash. If some of them are flaky, maybe we should re-think them to avoid it instead?

wayflowcore/tests/testhelpers/testhelpers.py
jschweiz: the previous code was pretty useless since if something doesn't get caught, it would raise anyway so the break would not be useful

wayflowcore/tests/utils.py
```diff
@@ -141,9 +141,78 @@ def _check_server_is_up(
     return False
 
 
-def get_available_port():
-    """Finds an available port to run a server"""
-    with closing(socket.socket(socket.AF_INET, socket.SOCK_STREAM)) as s:
-        s.bind(("", 0))
-        s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-        return s.getsockname()[1]
+def get_available_port(tmp_path: str):
```
jschweiz, line 144: could we have some unit test to check this behavior?

wayflowcore/setup.py
```diff
@@ -44,12 +44,13 @@ def read(file_name):
...
         "Programming Language :: Python :: 3.12",
         "Programming Language :: Python :: 3.13",
+        "Programming Language :: Python :: 3.14",
         "Topic :: Software Development :: Libraries :: Python Modules",
     ],
     keywords="NLP, text generation, code generation, LLM, Assistant, Tool, Agent",
     package_dir={"": "src"},
     packages=find_packages("src"),
-    python_requires=">=3.10,<3.14",
+    python_requires=">=3.10",
```
jschweiz: do we want to keep the upper bound to 3.15?

wayflowcore/tests/conftest.py
```diff
@@ -89,18 +89,15 @@
...
 
-import shutil
-import tempfile
-
-import pytest
-
 
 @pytest.fixture(scope="session")
-def session_tmp_path():
+def session_tmp_path(tmp_path_factory):
```
jschweiz, line 94: this returns a Path object, but it's typed as a str everywhere it's used, might be worth updating the typing

wayflowcore/tests/a2a/conftest.py
cesarebernardis: These tests are disabled, I'll remove it

wayflowcore/tests/test_docs_examples.py
cesarebernardis: From the `httpx` documentation https://www.python-httpx.org/api/#client:

> timeout - (optional) The timeout configuration to use when sending requests.

wayflowcore/tests/testhelpers/testhelpers.py
cesarebernardis: Yes, and python 3.14 now complains about that

wayflowcore/tests/run_tests.sh
cesarebernardis: But we do run it in parallel. I guess you mean running all the python versions in parallel, that's ok, I'll remove it as we are not increasing CI complexity in this PR.

wayflowcore/tests/integration/test_descriptors.py
cesarebernardis: The first line of the test where the `steps_to_check` list is used (which is the list of steps built using `step_cls_skiplist`) does:
```
if step_cls is Step:
    pytest.skip()
```
so I am skipping it here directly. Anyway, somehow the removal of these lines was not committed, I am doing it now.

wayflowcore/tests/conftest.py
cesarebernardis: Good catch, thanks

wayflowcore/tests/test_docs_examples.py
cesarebernardis: Increased to 5

wayflowcore/tests/test_docs_examples.py
cesarebernardis: We had a private discussion about this including @paul-cayet . We agree with what you are saying, but to avoid further flakiness in the CI, we decided to introduce this minimal retry for now (2 attempts, below the usual minimum of 3), and to tackle the documentation examples to make them more robust (see for example what I did in the `promptexecutionstep` how-to guide) in another PR, since it's out of scope for this one

wayflowcore/tests/utils.py
cesarebernardis: You mean removing `WAYFLOW_TEST`? I would rather keep it, as `port_span` might be too generic and potentially conflict with some other name

wayflowcore/tests/utils.py
cesarebernardis: Same as https://github.com/oracle/wayflow/pull/16/changes/BASE..68b534abacc5963f60277efef59bfee1d73e919e#r2726947426

wayflowcore/tests/utils.py
cesarebernardis: Added comment

wayflowcore/setup.py
cesarebernardis: I'll put it just in case

wayflowcore/src/wayflowcore/tools/toolhelpers.py
cesarebernardis: Added

wayflowcore/tests/utils.py
cesarebernardis: Added

PR #21, opened 2026-01-09
Support tracking of extra_content in OpenAI-compatible models that provide it

geeooleea:
Adds support for extra_content in tool requests to store additional content sent by OpenAI compatible APIs (e.g., gemini-3's  thought_signature).

Extra content is added as a private attribute of messages and tool requests. Tests are added to check for correct tracking in both scenarios.

### DIFF
--- a/docs/wayflowcore/source/core/changelog.rst
+++ b/docs/wayflowcore/source/core/changelog.rst
@@ -147,10 +147,10 @@ Improvements
 
   Added support for multiple levels of ManagerWorkers where a worker in ``ManagerWorkers`` can itself be ``ManagerWorkers`` enabling the creation of hierarchical agent structures.
 
-* **Added support for reasoning for certain OpenAI compatible models**
+* **Added support for Gemini reasoning models (via OpenAI compatible API)**
 
   The ``extra_content`` returned by Gemini's OpenAI-compatible APIs is now tracked
-  as part of messages and tool_requests. Gemini reasoning models are now supported.
+  as part of messages and tool_requests.
 
 
 Possibly Breaking Changes


nasseur:
Internal regression failed: Build ID #90

wayflowcore/tests/models/test_openaicompatiblemodel.py
```diff
@@ -626,3 +633,85 @@ def test_openai_model_does_not_raise_on_receiving_incomplete_tool_calls_from_rem
...
+
+
+@pytest.fixture
+def thought_signature_llm():
+    if "GEMINI_API_KEY" not in os.environ:
+        pytest.skip("Skipping test that requires access to a model with thought signatures")
+
+    return OpenAICompatibleModel(
+        model_id="gemini-3-pro-preview",
+        base_url="https://generativelanguage.googleapis.com/v1beta/openai/chat/completions",
```
sonleoracle, line 687: Did you try the v1 endpoint instead of v1beta?

wayflowcore/tests/models/test_openaicompatiblemodel.py
```diff
@@ -626,3 +633,85 @@ def test_openai_model_does_not_raise_on_receiving_incomplete_tool_calls_from_rem
...
+        api_key=os.environ["GEMINI_API_KEY"],
+        generation_config=LlmGenerationConfig(extra_args={"reasoning_effort": "low"}),
```
sonleoracle, line 689: Might be worth it to try [disabling reasoning](https://ai.google.dev/gemini-api/docs/openai#thinking) for 2.5 models:

>If you want to disable thinking, you can set reasoning_effort to "none" for 2.5 models. Reasoning cannot be turned off for Gemini 2.5 Pro or 3 models.

wayflowcore/src/wayflowcore/models/_openaihelpers/_chatcompletions_processor.py
```diff
@@ -248,25 +271,32 @@ async def _tagged_chunk_iterator_from_stream_of_openai_compatible_json(
...
-
+                    tool_requests_dict[index]["args"] += func["arguments"]
+            if "extra_content" in delta:
+                tool_requests_dict[index]["_extra_content"] = delta["extra_content"]
         return [
             ToolRequest(
                 name=s["name"],
                 tool_request_id=s["tool_request_id"],
-                args=_safe_json_loads(s["arguments"]),
+                args=_safe_json_loads(s["args"]),
```
sonleoracle, line 299: args instead of arguments?

wayflowcore/src/wayflowcore/models/_openaihelpers/_chatcompletions_processor.py
```diff
@@ -51,20 +64,24 @@ def _convert_message_into_openai_message_dict(self, m: "Message") -> List[Dict[s
...
+                        **(
+                            {"extra_content": tc._extra_content}
+                            if tc._extra_content is not None
+                            else {}
+                        ),
+                    }
+                    for tc in (m.tool_requests or [])
+                ],
+            }
+            return [openai_message]
```
sonleoracle: Nit: you could return the object immediately instead of assigning it to `openai_message`

sonleoracle (approved):

wayflowcore/tests/models/test_openaicompatiblemodel.py
geeooleea: There is no non-beta endpoint. [Official docs](https://ai.google.dev/gemini-api/docs/openai) still have beta in them, and just trying it gives 404.

wayflowcore/tests/models/test_openaicompatiblemodel.py
geeooleea: Yes that was actually how I first figured we needed to make these changes. Let me add a test for the non-reasoning models as well.

geeooleea: Added (posting screenshot because the test reqires API key):

grafik

wayflowcore/src/wayflowcore/models/_openaihelpers/_chatcompletions_processor.py
geeooleea: I created a new type `OpenAIToolRequestAsDictT` and renamed the key to args instead of arguments for consistency with the `ToolRequestAsDictT` (which I couldn't reuse because the "optionality" of the types is slightly different). Would you prefer to instead keep it consistent with the "arguments" coming from the openai apis?

sonleoracle: Ah ok I see, then I think your change is good

docs/wayflowcore/source/core/changelog.rst
```diff
@@ -107,6 +107,14 @@ Improvements
 
   This significantly reduces token usage and improves execution speed.
 
+
+* **Added support for reasoning for certain OpenAI compatible models**
+
+  The ``extra_content`` returned by some OpenAI-compatible APIs is now tracked
+  as part of messages and tool_requests. Some models may see improved performance
```
dhilloulinoracle: I think you can mention Gemini 3

nasseur:
Merge Gate failed: Build ID #2152

nasseur:
Merge Gate succeeded 🍏: Build ID #2154

PR #22, opened 2026-01-12
feat: multiple tool calling in Swarms

sathvikbhagavan:
Currently, Swarm does not support multiple tool calls, meaning when the llm returns multiple tool calls (in one message), only one tool call is handled, and the rest is cancelled.

This PR adds support for multiple tool calls by iterating over tool requests. Each iteration may start a new execution thread.

If the LLM returns multiple tool calls after `handoff_conversation`, any subsequent tool calls are ignored, since the conversation has already been transferred to another agent.

The Swarm prompt template is also modified to encourage the model to output multiple tool calls when appropriate.

Closes: #37

PR #23, opened 2026-01-12
feat: multiple tool calling in managerworkers

sathvikbhagavan:
This PR does:

- supports multiple tool calls for the manager agent in ManagerWorkers pattern (multiple tool calling for workers is already supported as it is a normal agent execution) by adding a "loop" to handle the pending tool requests of the manager explicitly.
- removes redundant parts in the ManagerWorkers's prompt template.
- modifies prompt to encourage multiple tool calls when appropriate.

This PR is similar to #22 

Closes: #38

### DIFF
--- a/wayflowcore/tests/test_swarm.py
+++ b/wayflowcore/tests/test_swarm.py
@@ -1333,25 +1333,7 @@ def test_swarm_can_do_multiple_tool_calling_when_appropriate(vllm_responses_llm)
     assert str(result) in conv.get_last_message().content
 
 
-@pytest.mark.parametrize(
-    "raise_exceptions, expected_exception, expected_result",
-    [
-        (True, pytest.raises(ValueError, match="Cannot compute result using fooza tool."), None),
-        (False, None, lambda: bwip_tool.func(4, 5) + zbuk_tool.func(5, 6)),
-    ],
-)
-@retry_test(max_attempts=4)
-def test_swarm_can_do_multiple_tool_calling_with_tool_raising_exception(
-    vllm_responses_llm, raise_exceptions, expected_exception, expected_result
-):
-    """
-    Failure rate:          1 out of 20
-    Observed on:           2026-01-16
-    Average success time:  17.15 seconds per successful attempt
-    Average failure time:  12.90 seconds per failed attempt
-    Max attempt:           4
-    Justification:         (0.09 ** 4) ~= 6.8 / 100'000
-    """
+def _setup_swarm_for_multiple_tool_calling(vllm_responses_llm, raise_exceptions):
     llm = vllm_responses_llm
     fooza_agent = _get_fooza_agent(
         llm, raise_exception_tool=True, raise_exceptions=raise_exceptions
@@ -1370,10 +1352,39 @@ def test_swarm_can_do_multiple_tool_calling_with_tool_raising_exception(
         messages="Compute the result of fooza(4, 2) + bwip(4, 5) + zbuk(5, 6)"
     )
 
-    if expected_exception:
-        with expected_exception:
-            conv.execute()
-    else:
+    return conv
+
+
+@retry_test(max_attempts=3)
+def test_swarm_can_do_multiple_tool_calling_with_tool_raising_exception_raises_error(
+    vllm_responses_llm,
+):
+    """
+    Failure rate:          0 out of 20
+    Observed on:           2026-01-28
+    Average success time:  5.19 seconds per successful attempt
+    Average failure time:  No time measurement
+    Max attempt:           3
+    Justification:         (0.05 ** 3) ~= 9.4 / 100'000
+    """
+    conv = _setup_swarm_for_multiple_tool_calling(vllm_responses_llm, raise_exceptions=True)
+    with pytest.raises(ValueError, match="Cannot compute result using fooza tool."):
         conv.execute()
-        result = expected_result()
-        assert str(result) in conv.get_last_message().content
+
+
+@retry_test(max_attempts=5)
+def test_swarm_can_do_multiple_tool_calling_with_tool_raising_exception_does_not_raise_error(
+    vllm_responses_llm,
+):
+    """
+    Failure rate:          2 out of 20
+    Observed on:           2026-01-28
+    Average success time:  14.45 seconds per successful attempt
+    Average failure time:  21.04 seconds per failed attempt
+    Max attempt:           5
+    Justification:         (0.14 ** 5) ~= 4.7 / 100'000
+    """
+    conv = _setup_swarm_for_multiple_tool_calling(vllm_responses_llm, raise_exceptions=False)
+    conv.execute()
+    result = bwip_tool.func(4, 5) + zbuk_tool.func(5, 6)
+    assert str(result) in conv.get_last_message().content


docs/wayflowcore/source/core/howtoguides/howto_serialize_conversations.rst
```diff
@@ -289,6 +289,6 @@ Full code
 
 Click on the card at the :ref:`top of this page <top-howtoserializeconversations>` to download the full code for this guide or copy the code below.
 
-.. literalinclude:: ../end_to_end_code_examples/tutorial_agent.py
+.. literalinclude:: ../end_to_end_code_examples/howto_serialize_conversations.py
```
paul-cayet, line 292: Nice catch, thanks

wayflowcore/src/wayflowcore/executors/_managerworkersexecutor.py
```diff
@@ -193,12 +211,63 @@ async def execute_async(
...
+
+            return (
+                _ToolProcessSignal.START_NEW_LOOP
+            )  # start a new loop to handle other pending tools if any
+
+        tool = next((t for t in manager.tools if t.name == tool_request.name), None)
+
+        # Case 2: standard client tool
+        if tool and isinstance(tool, ClientTool):
+            # Only one tool request is surfaced to the user at a time
```
paul-cayet: Please add a comment saying that the other tool calls are still in the tool queue (so that we don't think that here this is simply discarding the other tool calls)

wayflowcore/src/wayflowcore/executors/_managerworkersexecutor.py
```diff
@@ -116,6 +121,20 @@ async def execute_async(
...
+                    result = ManagerWorkersRunner._handle_pending_tool_requests_of_manager(
+                        manager=current_agent,
+                        manager_conversation=current_conversation,
+                        managerworkers_conversation=conversation,
+                    )
+
+                    if result == _ToolProcessSignal.RETURN:
+                        return current_conversation.status
+                    elif result == _ToolProcessSignal.START_NEW_LOOP:
+                        continue
```
paul-cayet, line 186: Can you add a 
```python
else:
    # _ToolProcessSignal.EXECUTE_AGENT -> we continue this loop with a new call to the current agent
    pass
```

wayflowcore/src/wayflowcore/executors/_managerworkersexecutor.py
```diff
@@ -214,18 +283,12 @@ def _send_message_to_manager(
...
-            )
-
-        last_tool_request_id = unanswered_tool_requests[0].tool_request_id
+        # Get the oldest unanswered tool requests since the tool requests are processed sequentially
+        tool_request_id = unanswered_tool_requests[0].tool_request_id
 
-        manager_subconversation.append_tool_result(
-            ToolResult(message.content, tool_request_id=last_tool_request_id)
+        # Manually append tool result directly to the message list (not via status._tool_results)
+        manager_subconversation.message_list.append_tool_result(
```
paul-cayet: Why do you need to change this? I would prefer if we don't change this

wayflowcore/tests/test_managerworkers.py
```diff
@@ -568,3 +577,257 @@ def test_managerworkers_execution_works_with_servertool_confirmation(big_llama):
...
+            multiple_tool_requests,  # fooza's output
+            "bwip answers to fooza",
+            "bwip answers to fooza",
+            "fooza answers to user",
+        ],
+    ):
+        status = conv.execute()
+        assert isinstance(status, UserMessageRequestStatus)
+        assert conv.get_last_message().content == "fooza answers to user"
+
```
paul-cayet, line 750: Please add a test to see what happens when a worker calls multiple tools in parallel including a tool that does not exist I want to make sure that this works as expected

wayflowcore/src/wayflowcore/executors/_managerworkersexecutor.py
```diff
@@ -193,12 +211,63 @@ async def execute_async(
...
+            return _ToolProcessSignal.EXECUTE_AGENT
+
+        # Case 1: send_message tool
+        if tool_request.name == _SEND_MESSAGE_TOOL_NAME:
+            ManagerWorkersRunner._send_message_to_worker(
+                managerworkers_conversation=managerworkers_conversation,
+                managerworkers_config=managerworkers_conversation.component,
+            )
+
+            manager_conversation.state.current_tool_request = None
```
paul-cayet: Please replace with manager_conv_state and same below

paul-cayet (approved):

wayflowcore/src/wayflowcore/executors/_managerworkersexecutor.py
sathvikbhagavan: Done in [6055391](https://github.com/oracle/wayflow/pull/23/commits/6055391d5e91c23112863eaf17c35b447f6cf104)

wayflowcore/src/wayflowcore/executors/_managerworkersexecutor.py
sathvikbhagavan: Yes its not needed, reverted in [6055391](https://github.com/oracle/wayflow/pull/23/commits/6055391d5e91c23112863eaf17c35b447f6cf104)

wayflowcore/src/wayflowcore/executors/_managerworkersexecutor.py
sathvikbhagavan: Done in [6055391](https://github.com/oracle/wayflow/pull/23/commits/6055391d5e91c23112863eaf17c35b447f6cf104)

wayflowcore/tests/test_managerworkers.py
sathvikbhagavan: Done in [43fe923](https://github.com/oracle/wayflow/pull/23/commits/43fe92344076dbe5ca3c947ddff89c96598dff95)

wayflowcore/src/wayflowcore/executors/_managerworkersexecutor.py
sathvikbhagavan: Done in [2065ab1](https://github.com/oracle/wayflow/pull/23/commits/2065ab1e85fefee3c75e44f5f5f2cfad2164c746)

wayflowcore/src/wayflowcore/templates/_managerworkerstemplate.py
```diff
@@ -36,30 +36,26 @@
 <other_entities>
```
jschweiz, line 36: can you check with tracing that the prompt is exactly what you expect? Make sure the new lines are correct, the prompt is well formed, we had too many problems like that in the past

wayflowcore/src/wayflowcore/templates/_managerworkerstemplate.py
```diff
@@ -82,14 +78,13 @@
...
 The user can only see the content of the messages sent with `talk_to_user` and will not see any of your thoughts.
--> Put **internal-only** information in the thoughts
--> Put all necessary information in the tool calls to communicate to user/other entity.
 
-Responses should be structured as a thought followed by a your response function call using JSON compliant syntax.
+Always structure your response as a thought followed by one or multiple tool calls using JSON compliant syntax.
 Do not use variables in the function call. Here's the structure:
 
-YOUR THOUGHTS (WHAT ACTION YOU ARE GOING TO TAKE; REMEMBER THAT THE USER CANNOT SEE THOSE!)
+YOUR THOUGHTS (WHAT ACTION YOU ARE GOING TO TAKE; NOT VISIBLE TO THE USER)
```
jschweiz: how did you measure this works better than the previous one? On what model?

wayflowcore/tests/test_managerworkers.py
```diff
@@ -568,3 +577,318 @@ def test_managerworkers_execution_works_with_servertool_confirmation(big_llama):
...
+@pytest.mark.parametrize(
+    "raise_exceptions, expected_exception, expected_result",
+    [
+        (True, pytest.raises(ValueError, match="Cannot compute result using fooza tool."), None),
+        (False, None, lambda: bwip_tool.func(4, 5) + zbuk_tool.func(5, 6)),
+    ],
+)
+@retry_test(max_attempts=4)
+def test_managerworkers_can_do_multiple_tool_calling_with_tool_raising_exception(
```
jschweiz: please split this test in two, because it checks 2 different behaviors: 1 works without raising, 1 raises an exception. You can use a fixture to setup both tests with the same agent, this way no need for parametrize

wayflowcore/tests/test_managerworkers.py
```diff
@@ -568,3 +577,318 @@ def test_managerworkers_execution_works_with_servertool_confirmation(big_llama):
...
+        outputs=[
+            multiple_tool_requests,  # fooza's output
+            "bwip answers to fooza",
+            "bwip answers to fooza",
+            "fooza answers to user",
+        ],
+    ):
+        status = conv.execute()
+        assert isinstance(status, UserMessageRequestStatus)
+        assert conv.get_last_message().content == "fooza answers to user"
```
jschweiz: check the list of messages is like you expect

wayflowcore/tests/test_managerworkers.py
```diff
@@ -568,3 +577,318 @@ def test_managerworkers_execution_works_with_servertool_confirmation(big_llama):
...
+                "message": "calculate bwip(4,5)",
+            },
+        ),
+    ]
+    with patch_llm(
+        llm,
+        outputs=[
+            multiple_tool_requests,  # fooza's output
+            "bwip answers to fooza",
+            "bwip answers to fooza",
```
jschweiz: nit:
```suggestion
            "bwip answers to second fooza call",
```
for clarity

wayflowcore/src/wayflowcore/executors/_managerworkersexecutor.py
```diff
@@ -116,6 +121,23 @@ async def execute_async(
 
             if current_agent_name == managerworkers_config.manager_agent.name:
                 current_agent = managerworkers_config.manager_agent
+
+                # Handle pending tool requests of manager
+                if isinstance(current_conversation.status, ToolRequestStatus):
+                    result = ManagerWorkersRunner._handle_pending_tool_requests_of_manager(
```
jschweiz: nitnit:
```suggestion
                    signal = ManagerWorkersRunner._handle_pending_tool_requests_of_manager(
```

wayflowcore/src/wayflowcore/executors/_managerworkersexecutor.py
```diff
@@ -193,12 +214,63 @@ async def execute_async(
...
+            return (
+                _ToolProcessSignal.START_NEW_LOOP
+            )  # start a new loop to handle other pending tools if any
+
+        tool = next((t for t in manager.tools if t.name == tool_request.name), None)
+
+        # Case 2: standard client tool
+        if tool and isinstance(tool, ClientTool):
+            # Only one tool request is surfaced to the user at a time, other tool calls are still in the queue
+            manager_conversation.status.tool_requests = [tool_request]
```
jschweiz, line 285: shouldn't you instead assign a new status to the conversation?

jschweiz: why are we using an enum, can't we return in this function a status, and in the caller act depending on this status?

wayflowcore/src/wayflowcore/executors/_managerworkersexecutor.py
```diff
@@ -116,6 +121,23 @@ async def execute_async(
 
             if current_agent_name == managerworkers_config.manager_agent.name:
```
jschweiz, line 172: can we extract all this code related to the manager in a specific method, to simplify this function? This way, this function will be much simpler:
```
while True:
   if manager:
      run_one_manager_round()
   else:
      run_one_worker_round()
   # case distinction to return to find what is next
```

wayflowcore/src/wayflowcore/executors/_managerworkersexecutor.py
sathvikbhagavan: > why are we using an enum, can't we return in this function a status, and in the caller act depending on this status?

It makes it easier to check if there are more tool requests remaining which needs to be executed or its a client tool to yield back to the user.

wayflowcore/tests/test_managerworkers.py
sathvikbhagavan: Yes, I combined this two as the tool is raising exceptions in both cases but in one test, the exception is propagated and in other, the LLM receives the error message and computes the partial result. So, can't it remain combined in the sense it is testing two ways to deal with exception raising tools? Maybe I can rename the test to something better?

wayflowcore/src/wayflowcore/templates/_managerworkerstemplate.py
sathvikbhagavan: I didn't make this change. Tbh, I don't think it would be too different. I will revert this change.

sathvikbhagavan: Removed it in [18bb461](https://github.com/oracle/wayflow/pull/23/commits/18bb461ab0ca037b0662a81cdedda478507e24a4)

wayflowcore/tests/test_managerworkers.py
sathvikbhagavan: Done in [029ae20](https://github.com/oracle/wayflow/pull/23/commits/029ae2026ae83a04326350db24180accba954f29)

wayflowcore/tests/test_managerworkers.py
sathvikbhagavan: Done in [029ae20](https://github.com/oracle/wayflow/pull/23/commits/029ae2026ae83a04326350db24180accba954f29)

wayflowcore/src/wayflowcore/executors/_managerworkersexecutor.py
sathvikbhagavan: Done in [f143e7a](https://github.com/oracle/wayflow/pull/23/commits/f143e7a027ceceae12a78c6cd07ffbc7f13b0b2d)

wayflowcore/src/wayflowcore/executors/_managerworkersexecutor.py
sathvikbhagavan: Yes. Done in [f143e7a](https://github.com/oracle/wayflow/pull/23/commits/f143e7a027ceceae12a78c6cd07ffbc7f13b0b2d)

wayflowcore/src/wayflowcore/templates/_managerworkerstemplate.py
sathvikbhagavan: Yes, it looks correct
image

wayflowcore/src/wayflowcore/executors/_managerworkersexecutor.py
sathvikbhagavan: I am resolving this as we discussed this offline.

wayflowcore/tests/test_managerworkers.py
sathvikbhagavan: I have split it in [e6154b6](https://github.com/oracle/wayflow/pull/23/commits/e6154b647c683e211e72a05c6f399cc8e827a2da)

nasseur:
Merge Gate INPROGRESS

PR #34, opened 2026-01-16
Upgrade Agent Spec, MCP, and Update Third-Party Licenses

younes-makhchan:
#### Changes
- Upgrade pyagentspec to 26.1.0
- Upgrade MCP to 1.24.0
- Update the third-party licenses file to reflect the new dependencies


aadnaane (approved):

wayflowcore/setup.py
```diff
@@ -51,7 +51,7 @@ def read(file_name):
     packages=find_packages("src"),
     python_requires=">=3.10,<3.14",
     install_requires=[
-        "pyagentspec>=25.4.1",
+        "pyagentspec>=26.1.0",
```
ElayachiFouad, line 54: this will prevent you from installing the dev version in the CI. Maybe we need to update the VERSION file to 26.2.0.dev0 and then generate a new release so it can be used in the CI

younes-makhchan: Since we are relying on the constraint_dev.txt for pyagentspec 26.1.0.dev5, we can just update it to 26.1.0. and they will use official release version

dhilloulinoracle (approved):

THIRD_PARTY_LICENSES.txt
```diff
@@ -1,18 +1,20 @@
...
 ========================================
+### Deprecated
+
+Provider: Laurent LAPORTE
 
-MIT
+Component(s): Deprecated
 
-Details:
+```
```
paul-cayet: Is it expected to have those triple backticks? It shows a bit weird on the text file https://github.com/oracle/wayflow/blob/e0dadc33b304fa9d66d247d27617cfa3b2ff779a/THIRD_PARTY_LICENSES.txt

younes-makhchan: yes, make sense, i will remove them

PR #39, opened 2026-01-19
docs: conceptual guide on data flow edges

paul-cayet:
This PR adds a guide to better understand how Data Flow Edges work and when they should be added to flows


docs/wayflowcore/source/core/_static/conceptual/dataflowedges_looping.svg
cesarebernardis, line 1: The top image says `Flow is incorrect`, and it makes it look like there are 2 errors, but the fact that nobody is consuming output D is not really an issue. Maybe we can be a bit more clear on what is the issue for D there.

The bottom image has renamed inputs and outputs, while I would expect the same letters I see in the top image, also considering the code that is linked to it.

docs/wayflowcore/source/core/code_examples/conceptual_dataflowedges.py
```diff
@@ -0,0 +1,128 @@
+# Copyright © 2025 Oracle and/or its affiliates.
```
cesarebernardis, line 1: Nit: Happy new year 🥳

docs/wayflowcore/source/core/code_examples/conceptual_dataflowedges.py
```diff
@@ -0,0 +1,128 @@
...
+# fmt: off
+# mypy: ignore-errors
+# docs-title: Code Example - Data Flow Edges in Flows
+
+# .. start-##_Flow_with_multi_output_routing
+from wayflowcore.flow import Flow
+from wayflowcore.steps import OutputMessageStep
+from wayflowcore.controlconnection import ControlFlowEdge
+
+# Flow with one step ouptut used by two subsequent steps
```
cesarebernardis:
```suggestion
# Flow with one step output used by two subsequent steps
```

docs/wayflowcore/source/core/code_examples/conceptual_dataflowedges.py
```diff
@@ -0,0 +1,128 @@
...
+# .. end-##_Flow_with_multi_output_routing
+conv = flow.start_conversation()
+_ = conv.execute()
+# .. start-##_Flow_with_multi_output_routing_with_explicit_edges
+from wayflowcore.flow import Flow
+from wayflowcore.steps import OutputMessageStep
+from wayflowcore.controlconnection import ControlFlowEdge
+from wayflowcore.dataconnection import DataFlowEdge
+
+producer = OutputMessageStep(name="producer", message_template="value")
```
cesarebernardis: Being a bit nitpicky, this is not exactly the scenario shown in the image associated to it (dataflowedges_routing.svg). In the image, the `producer` outputs `A`, and `consumer1` has an input `C` (or `A`, but apparently only after renaming), while here `producer` outputs another name and `consumer1` has input `A` already. I would try to align code and image.

docs/wayflowcore/source/core/conceptual_guides/data_flow_edges.rst
```diff
@@ -0,0 +1,162 @@
+.. _conceptual_dataflowedges:
```
cesarebernardis, line 1: Are we going to have a similar guide also in Agent Spec? I think it would fit there as well.

docs/wayflowcore/source/core/conceptual_guides/data_flow_edges.rst
```diff
@@ -0,0 +1,162 @@
...
+
+.. image:: ../_static/conceptual/dataflowedges_routing.svg
+    :align: center
+    :scale: 100%
+    :alt: Data Flow Edges for a Flow with data routing
+
+
+In this first example, Step 1 produces an output that needs to be sent to two different steps: Step 2 and Step 3.
+
+* When the name of the value is different between the steps (e.g., "A" and "B"), WayFlow cannot automatically
```
cesarebernardis, line 77: Nit: align names of steps and I/Os to the ones in the image

nasseur:
Internal regression failed: Build ID #237

docs/wayflowcore/source/core/code_examples/conceptual_dataflowedges.py
paul-cayet: Prefer to do bulk update

docs/wayflowcore/source/core/_static/conceptual/dataflowedges_looping.svg
paul-cayet: The error is because the inputs/outputs are incorrectly named so it is necessary to keep them as is. I will make the text more explicit

docs/wayflowcore/source/core/index.rst
```diff
@@ -109,6 +109,7 @@ Look through our :doc:`Frequently Asked Questions <faqs>`.
 
    Tutorials & Use Cases <tutorials/index>
    How-to Guides <howtoguides/index>
+   Conceptual Guides <conceptual_guides/data_flow_edges>
```
jschweiz, line 112: no index? Don't we plan to add more guides?

docs/wayflowcore/source/core/conceptual_guides/data_flow_edges.rst
```diff
@@ -0,0 +1,162 @@
...
+    .. tab:: Code with explicit data routing
+
+        .. literalinclude:: ../code_examples/conceptual_dataflowedges.py
+            :language: python
+            :start-after: .. start-##_Flow_with_multi_output_routing_with_explicit_edges
+            :end-before: .. end-##_Flow_with_multi_output_routing_with_explicit_edges
+
+
+
+Example 2: Looping Flows
```
jschweiz, line 107: it's worth adding a warning that mentions why input A needs to be optional (in the first loop, it hasn't produced a value yet)

jschweiz (approved):

docs/wayflowcore/source/core/index.rst
paul-cayet: Will add later when we have more than one guide

docs/wayflowcore/source/core/conceptual_guides/data_flow_edges.rst
paul-cayet: Yes indeed, will add it

Issue #37, opened 2026-01-19
Add handling of multiple tool requests in Swarms


sathvikbhagavan:
Currently, in Swarms, if an agent generates multiple tool requests, only one of them is executed and rest of them are cancelled. We can extend the behavior to handle multiple tool requests.

PR #44, opened 2026-01-20
feat: multiple tool calling in Swarms 

sathvikbhagavan:
Currently, Swarm does not support multiple tool calls, meaning when the llm returns multiple tool calls (in one message), only one tool call is handled, and the rest is cancelled.

This PR adds support for multiple tool calls by iterating over tool requests. Each iteration may start a new execution thread.

If the LLM returns multiple tool calls after handoff_conversation, any subsequent tool calls are ignored, since the conversation has already been transferred to another agent.

The Swarm prompt template is also modified to encourage the model to output multiple tool calls when appropriate.

Closes: https://github.com/oracle/wayflow/issues/37

Duplicate of https://github.com/oracle/wayflow/pull/22, have some issues with comments resolution on github side, hence creating a new PR.


nasseur:
Internal regression succeeded 🍏: Build ID #1425

PR #45, opened 2026-01-20
Support Agent Spec Sensitive Headers

cesarebernardis:
Introduce the sensitive headers from Agent Spec, and add their conversion.

Sensitive headers are a set of headers that might contain sensitive information. These headers are excluded from any serialization for security reasons.

### DIFF
--- a/wayflowcore/tests/agentspec/test_sensitive_fields.py
+++ b/wayflowcore/tests/agentspec/test_sensitive_fields.py
@@ -10,11 +10,7 @@ import pytest
 
 from wayflowcore.agent import Agent
 from wayflowcore.agentspec import AgentSpecExporter, AgentSpecLoader
-from wayflowcore.mcp import (
-    MCPToolBox,
-    SSEmTLSTransport,
-    StreamableHTTPmTLSTransport,
-)
+from wayflowcore.mcp import MCPToolBox, SSEmTLSTransport, StreamableHTTPmTLSTransport
 from wayflowcore.models import OCIGenAIModel, OpenAICompatibleModel
 from wayflowcore.models.ociclientconfig import OCIClientConfigWithApiKey
 from wayflowcore.steps import ApiCallStep


wayflowcore/src/wayflowcore/mcp/clienttransport.py
```diff
@@ -154,6 +159,10 @@ class RemoteBaseTransport(SerializableDataclass, ClientTransport, ABC):
     headers: Optional[Dict[str, str]] = None
     """The headers to send to the server."""
 
+    sensitive_headers: Optional[Dict[str, str]] = None
+    """The sensitive headers to send to the server, joined with the ones defined in the ``headers``.
+    They will be excluded from any serialization for security reasons."""
```
lfaucon: minor: Added this validation in pyagentspec (maybe not everywhere), the same would be nice here

```
    @model_validator_with_error_accumulation
    def _validate_sensitive_headers_are_disjoint(self) -> Self:
        repeated_headers = set(self.headers or {}).intersection(set(self.sensitive_headers or {}))
        if repeated_headers:
            raise ValueError(
                f"Found some headers have been specified in both `headers` and "
                f"`sensitive_headers`: {repeated_headers}"
            )
        return self
```

wayflowcore/src/wayflowcore/mcp/clienttransport.py
```diff
@@ -249,7 +277,7 @@ class SSETransport(RemoteBaseTransport, ClientTransportWithAuth, SerializableObj
     def _get_client_transport_cm(self) -> ClientTransportContextManagerType:
         return sse_client(
             url=self.url,
-            headers=self.headers,
+            headers=self._merge_headers(),
```
lfaucon: minor/codestyle: the method name `_merge_headers` is a command verb, but the method is actually a query https://en.wikipedia.org/wiki/Command%E2%80%93query_separation

`_get_merged_headers` or make it a property called `_merged_headers`

wayflowcore/tests/agentspec/test_sensitive_fields.py
```diff
@@ -0,0 +1,155 @@
...
+    assert mcp_toolbox.client_transport.ssl_ca_cert == "path/to/abcdexyz.pem"
+
+
+def test_import_component_with_incomplete_missing_sensitive_fields_fails(
+    agent_with_sensitive_fields_filepath: str,
+) -> None:
+    incomplete_components_registry = {
+        "oci-client-config.auth_file_location": "path/to/abcdexyz.json",
+        "remote-tool-id.sensitive_headers": {"Authorization": "Bearer abcdexyz"},
+        # "mcp-toolbox-client-transport-id.key_file": "path/to/abcdexyz.json",
```
jschweiz, line 152: why is that commented? If it needs to be commented, add a comment to mention why

wayflowcore/tests/serialization/test_tool_serialization.py
```diff
@@ -228,13 +231,16 @@ def test_remote_tool_can_be_serialized(remote_tool):
     serialized_tool = serialize(remote_tool)
     assert "my_api" in serialized_tool
     assert "POST" in serialized_tool
+    assert "sensitive_headers" not in serialized_tool
```
jschweiz, line 234: I guess it's fine for now, since we don't support it anywhere else, but now we should update WayFlow to still export the ID of the Sensitive attribute, so we can re-inject it when we import again

wayflowcore/src/wayflowcore/mcp/clienttransport.py
```diff
@@ -171,6 +180,25 @@ class RemoteBaseTransport(SerializableDataclass, ClientTransport, ABC):
     session_parameters: SessionParameters = field(default_factory=SessionParameters)
     """Arguments for the MCP session."""
 
+    def _merge_headers(self) -> Optional[Dict[str, str]]:
+        # Merge sensitive and non-sensitive headers (if any) in a single dictionary
+        merged_headers: Dict[str, str] = {}
+        if self.headers:
```
jschweiz, line 196: if there is a conflict during merge, what happens?

wayflowcore/src/wayflowcore/tools/remotetools.py
```diff
@@ -284,3 +300,14 @@ def __init__(
     @property
     def _tool_type(self) -> SupportedToolTypesT:
         return "remote"
+
+    def _serialize_to_dict(self, serialization_context: "SerializationContext") -> Dict[str, Any]:
```
lfaucon, line 306: major: Is that legacy serialization now broken in the sense that if any component has sensitive headers then there is no way to deserialize correctly? Would it be better to raise here when `sensitive_headers` are present rather than successfully doing something wrong?

wayflowcore/src/wayflowcore/mcp/clienttransport.py
dhilloulinoracle: the code show that the sensitive headers take precedence.
Might be good to document though

wayflowcore/tests/mcptools/test_mcp_tools.py
```diff
@@ -503,6 +521,23 @@ def test_serde_mcp_tool(mcp_fooza_tool):
...
+    assert isinstance(deserialized_tool, MCPTool)
+    assert isinstance(deserialized_tool.client_transport, SSETransport)
+    assert type(mcp_fooza_tool_with_client_with_headers.client_transport) is type(
+        deserialized_tool.client_transport
+    )
+    assert (
+        deserialized_tool.client_transport.headers
+        == mcp_fooza_tool_with_client_with_headers.client_transport.headers
+    )
+    assert deserialized_tool.run(a=1, b=2) == "7"
```
lfaucon, line 550: major (same concern as https://github.com/oracle/wayflow/pull/45#discussion_r2726922099):
In practice, if the sensitive header is not ignored as in this test, then it would fail the execution, no? Isn't it a problem?

wayflowcore/src/wayflowcore/mcp/clienttransport.py
lfaucon: We went for the approach of raising validation errors on conflicts for these in pyagentspec

wayflowcore/src/wayflowcore/tools/remotetools.py
dhilloulinoracle: I don't think we should throw. We should ideally change the deserialization so that it is possible to add the sensitive things, similar to how we do for agent spec ser/des

wayflowcore/tests/serialization/test_tool_serialization.py
cesarebernardis: Yes, as long as we want to continue supporting wayflow's serialization separately from agent spec one, we should introduce a mechanism to do this (and other parts will benefit, like DB sensitive attributes, api keys, etc.)

wayflowcore/tests/agentspec/test_sensitive_fields.py
cesarebernardis: Commented

wayflowcore/src/wayflowcore/mcp/clienttransport.py
cesarebernardis: Went for the property alternative

wayflowcore/src/wayflowcore/tools/remotetools.py
cesarebernardis: As mentioned in https://github.com/oracle/wayflow/pull/45#discussion_r2730973086, we already have this issue with other attributes, and if we want to keep wayflow's serialization, we should introduce a "disaggregated components" mechanism similar to the agent spec one.

I would say that the change in this PR is backward compatible, since old serializations cannot contain `sensitive_headers`, as it did not exist, and the deserialization of `headers` did not change. From now on, users are aware that sensitive headers are not serialized.

wayflowcore/tests/mcptools/test_mcp_tools.py
cesarebernardis: > then it would fail the execution, no?

Yes

> Isn't it a problem?

I would say it is not, or at least users should be aware of this behavior. We say in several places that sensitive headers are not serialized, and we even log a warning if users serialize something that contains non-empty sensitive headers. To support this scenario properly, we should introduce in wayflow's serialization a "disaggregated mechanism" similar to agent spec, but that's out of the scope of this PR.

wayflowcore/src/wayflowcore/mcp/clienttransport.py
cesarebernardis: Changed logic to raise an error if `headers` and `sensitive_headers` keys overlap

wayflowcore/src/wayflowcore/mcp/clienttransport.py
cesarebernardis: Added

wayflowcore/tests/mcptools/test_mcp_tools.py
lfaucon: Ok, that makes sense

docs/wayflowcore/source/core/changelog.rst
```diff
@@ -1,6 +1,25 @@
...
+New features
+^^^^^^^^^^^^
+
+Improvements
+^^^^^^^^^^^^
+
+* **Added ``sensitive_headers`` in components that perform remote calls:**
+
+  ``ApiCallStep``, ``RemoteTool``, and MCP ``RemoteBaseTransport`` now have a new attribute ``sensitive_headers``.
+  This new attribute should be used for headers that might contain sensitive information (e.g., authentication).
```
dhilloulinoracle: @cesarebernardis , can we put it in 26.1?

cesarebernardis: Yes, moved to 26.1 changelog

nasseur:
Internal regression failed: Build ID #1455

PR #46, opened 2026-01-20
Support mcp text resources in tools that do not define their output descriptors

jschweiz:
Tools that do not define their output descriptors default to `string`, which does not handle when the actual tool result is a mcp resrouce. We need to take care of that since many MCP servers (like the github one) actually return resources without specifying the tool outputs.

Solution:
- when the MCP tool does not specify its outputs, we default to string and we will stringify any text or resource so that the LLM can access it

### DIFF
--- a/wayflowcore/tests/mcptools/test_mcp_tools.py
+++ b/wayflowcore/tests/mcptools/test_mcp_tools.py
@@ -98,7 +98,7 @@ def streamablehttp_client_transport_mtls(
 def run_toolbox_test(transport: ClientTransport) -> None:
     toolbox = MCPToolBox(client_transport=transport)
     tools = toolbox.get_tools()  # need
-    assert len(tools) == 10
+    assert len(tools) == 11
     mcp_tool = next(t for t in tools if t.name == "fooza_tool")
     assert mcp_tool.run(a=1, b=2) == "7"
     assert mcp_tool.input_descriptors == [IntegerProperty(name="a"), IntegerProperty(name="b")]


dhilloulinoracle:
@jschweiz please described in the PR description the solution you went with.
Do we end up supporting "resources"?

wayflowcore/src/wayflowcore/mcp/mcphelpers.py
```diff
@@ -98,8 +98,20 @@ def _extract_text_content_from_tool_result(result: types.CallToolResult) -> str:
...
+        elif content.type == "resource" and isinstance(
+            content.resource, types.TextResourceContents
+        ):
+            resource = content.resource
+            content_as_json = {
+                "uri": str(resource.uri),
+                "mimeType": resource.mimeType,
+                "text": resource.text,
+            }
+            text_content += json.dumps(content_as_json)
```
dhilloulinoracle, line 110: my understanding is that text resources might be quite large?
So we will just return them anyway, and potentially blow up the LLM context next turn?
Can we ignore some outputs (not return them out) if the developer specifies outputs to keep?

dhilloulinoracle: or developers can only declare the outputs to precise their types but they all have to be returned?

jschweiz: it's easy to control thebehavior of `MCPTool` since we can define the outputs, but in my case I'm using a `MCPToolBox`, for which I can't easily set the outputs of a particular function ^^ So I need to trust that the content returned by any tool can be visible by the LLM.
For EmbeddedResource, the content is passed as an output of the tool, so I think it's fine to put it in the context of the LLM. If the resource is too large, its content would probably not be returned here, just the URI, and then when needed the agentic system can fetch the actual content of the resource using the URI

paul-cayet: Maybe it would be nice to be able to have some way to cap the max size of an output from a tool in general though, similar to the `after_tool_callback` from Google ADK https://google.github.io/adk-docs/callbacks/

wayflowcore/src/wayflowcore/mcp/mcphelpers.py
```diff
@@ -98,8 +98,20 @@ def _extract_text_content_from_tool_result(result: types.CallToolResult) -> str:
     for content in result.content:
         if content.type == "text":
             text_content += content.text
+        elif content.type == "resource" and isinstance(
+            content.resource, types.TextResourceContents
+        ):
+            resource = content.resource
+            content_as_json = {
+                "uri": str(resource.uri),
```
sonleoracle, line 106: Why does the LLM need to see this?

jschweiz: in case it wants to read it later? Or if the content is not attached here and the LLM needs to go fetch it separately

nasseur:
Internal regression failed: Build ID #336

PR #49, opened 2026-01-21
Add llama4 pythonic template

jschweiz:
Add default llama4 template to prevent llama4 vllm tool calling issues


nasseur:
Internal regression failed: Build ID #289

PR #54, opened 2026-01-23
fix: make agent transforms run before template transforms to avoid bugs with CanonicalizationMessageTransform

youssef62:
**Context** 
Canonicalization was merging consecutive user messages before summarization could trigger, preventing proper conversation summarization when using Gemma models. You can see this in the `test_summarization_transforms.py::test_agent_transforms_should_run_before_canonicalization_with_gemma`. 

So now we run the agent transforms before all the default transforms inside the templates to avoid these types of problems. In this example,  `CanonicalizationMessageTransform` will run **after**  `ConversationSummarizationTransform`.


wayflowcore/src/wayflowcore/models/vllmmodel.py
```diff
@@ -149,20 +149,12 @@ def default_chat_template(self) -> "PromptTemplate":
...
                 "Llama-3.x models have limited performance with native tool calling. Wayflow will instead use the `LLAMA_AGENT_TEMPLATE`, which yields better performance than native tool calling"
             )
             return LLAMA_AGENT_TEMPLATE
-        if _is_gemma_model(self.model_id):
-            logger.debug(
-                "Gemma models only support alternating user and agent messages. The `CanonicalizationMessageTransform` will be added to this default's model template to ensure that."
-            )
-            return NATIVE_AGENT_TEMPLATE.with_additional_post_rendering_transform(
-                CanonicalizationMessageTransform()
-            )
```
lfaucon, line 166: The code change seems a bit complicated. The new implementation causes the agent implementation to have to know about details of Llm Models, while we would prefer to keep the knowledge of specific models localized to the models' implementation (you see this as you have to add the import statement `from wayflowcore.models._modelhelpers import _is_gemma_model` in the agents submodule).

A different fix that I think would also do the job well would be that `with_additional_post_rendering_transform` accepts a new argument `prepend: bool = False` that is set to True when the agent adds its own transforms.

What do you think?

youssef62: I agree that what you suggest is much better. It is now the implemented method.

nasseur:
Internal regression succeeded 🍏: Build ID #460

wayflowcore/tests/transforms/test_summarization_transforms.py
```diff
@@ -627,7 +633,9 @@ def test_summarization_transform_removes_expired_messages(
 
     cache_key = ""
     if transform_type == MessageSummarizationTransform:
-        cache_key = str(first_conv.id) + "_1_content"  # second message is the long one.
+        cache_key = (
+            str(first_conv.id) + "_2_content"
+        )  # second message in the list is the long one. Which is thrid overall (system prompt)
```
dhilloulinoracle: typo

wayflowcore/src/wayflowcore/agent.py
```diff
@@ -311,8 +311,10 @@ def __init__(
 
         self.transforms = transforms or []
         agent_template = agent_template or llm.agent_template
-        for transform in self.transforms:
-            agent_template = agent_template.with_additional_post_rendering_transform(transform)
+        for transform in self.transforms[::-1]:
```
dhilloulinoracle, line 326: please put comments about why it is done this way

wayflowcore/src/wayflowcore/agent.py
```diff
@@ -311,8 +311,10 @@ def __init__(
...
+            agent_template = agent_template.with_additional_post_rendering_transform(
+                transform, append=False
```
dhilloulinoracle: explain why it needs append=False

dhilloulinoracle: the name "append" is pretty poor.
I looked at the code, and it means "append_last" or something like that.
Please add that "append_last" parameter, deprecate the previous "append" one. (To deprecate, both add a code comment indicating when it was deprecated and when it will be removed - wayflow 26.3 - and add a deprecate notice in the changelog)
There are other methods in the same file where we should do that as well.

wayflowcore/src/wayflowcore/agent.py
```diff
@@ -740,6 +745,25 @@ def _convert_described_flow_into_named_flow(described_flow: Union[DescribedFlow,
...
+            isinstance(transform, CanonicalizationMessageTransform)
+            for transform in (template.post_rendering_transforms or [])
+        )
+        if not has_canonicalization:
+            logger.debug(
+                "Gemma models only support alternating user and agent messages. The `CanonicalizationMessageTransform` will be added when creating an agent."
+            )
+            return template.with_additional_post_rendering_transform(
+                CanonicalizationMessageTransform()
+            )
```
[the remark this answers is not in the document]
youssef62: After @lfaucon's comment, we changed the implementation, so this comment is not relevant anymore.

wayflowcore/src/wayflowcore/agent.py
youssef62: I did this in [8395b82](https://github.com/oracle/wayflow/pull/54/commits/8395b82d39bd52b82980e0f23aca15420ca9a023). Let me know If I deprecated correctly. 

Also, I did not find `append` to be really confusing, I find the current situation with both `append` and `append_last` slightly more confusing. But this is highly subjective, let me know if you want me to keep or remove the commit.

wayflowcore/src/wayflowcore/agent.py
youssef62: Done in [57943bf](https://github.com/oracle/wayflow/pull/54/commits/57943bfea347ab8873602d50399da3566dc78a2d).

wayflowcore/tests/transforms/test_summarization_transforms.py
youssef62: Thanks for catching it. Corrected in [445354f](https://github.com/oracle/wayflow/pull/54/commits/445354fc2a5d5f2ba370bd94cae27aa13349a3df)

wayflowcore/src/wayflowcore/agent.py
dhilloulinoracle: thanks, I do think that it will be better once we remove the `append` one.
It doesn't seem right to have something `with_additional_bla(append=False)`

wayflowcore/src/wayflowcore/agent.py
```diff
@@ -311,8 +311,14 @@ def __init__(
...
         self.transforms = transforms or []
         agent_template = agent_template or llm.agent_template
-        for transform in self.transforms:
-            agent_template = agent_template.with_additional_post_rendering_transform(transform)
+
+        # Iterate in reverse order to keep the original order of transforms as we are prepending.
+        for transform in self.transforms[::-1]:
+            # Prepend to make sure that user defined transforms are applied before default template transforms.
+            # This ensures a more predictable outcome for users.
+            agent_template = agent_template.with_additional_post_rendering_transform(
```
jschweiz: agent transforms are pre-rendering transforms, not post_rendering transforms... I think this  is the bug that's causing problem, no?
Summarization should definitely be a pre-rendering, because we don't want the summarization to mess with LLM formatting. 
`pre-rendering-transform`: change the context of the LLM
`post-rendering-tranform`: change the format of what is passed to the LLM

lfaucon: > Summarization should definitely be a pre-rendering

🤔 We concluded before that summarization is definitely a post-rendering transform

Because pre-rendering a message would look like `"Something something {{long_document_placeholder}}"` so summarization at pre-rendering would (1) risk breaking placeholders (2) not summarize content coming from input variables.

Am I understanding this wrong?

docs/wayflowcore/source/core/changelog.rst
```diff
@@ -145,6 +132,11 @@ Possibly Breaking Changes
 
   - For now, a deprecation warning is raised; in the next cycle, an error will be thrown.
 
+* **Deprecated 'append' parameter in PromptTemplate transform methods:**
+
+  The ``append`` parameter in ``with_additional_post_rendering_transform`` and ``with_additional_pre_rendering_transform`` methods is deprecated.
+  Use ``append_last`` instead. The ``append`` parameter will be removed in WayFlow 26.3.
```
lfaucon, line 184: You can see this separate PR adding the new changelog version: https://github.com/oracle/wayflow/pull/45/changes#diff-5763de0479f70d3baed2f7fa1b8511266eb6dfad75a5116acf3f8a04a5d1aefdR4

I don't think it is correct for this PR to update the changelog of version 26.1.0

wayflowcore/src/wayflowcore/agent.py
jschweiz: Mmh but placeholders are only in the system prompt, which is injected just before the post rendering transforms. You probably never want to summarize the system  prompt anyway, so it should not have any impact (so you can't break placeholders and you're never summarizing content coming from input variables)

In my mind, summarization is to condense the past, and the past is the chat history, which is the pre-rendering. If you summarize post-rendering, imagine the post-rendering transforms are not in the right order, the one responsible of changing the conversation format (using react, or pythonic calls, ...) will then be overriden by summaries?

lfaucon: Ok. I understand & agree 👍 Thanks for the clarification.

youssef62: I agree too with this! I added it in [fa85948](https://github.com/oracle/wayflow/pull/54/commits/fa859488e087478cae866b38e02777fce9c18f4e). 

I still do `append_last=False` and reverse order loop in order to still have the agent transforms execute before potential default  **pre**_rendering_transforms (which now don't exist).

Let me know if you approve of this.

nasseur:
Internal regression succeeded 🍏: Build ID #487

docs/wayflowcore/source/core/changelog.rst
youssef62: (we discussed this offline and we decided to keep it part of 26.1.0).

wayflowcore/src/wayflowcore/agent.py
```diff
@@ -311,8 +311,14 @@ def __init__(
 
         self.transforms = transforms or []
         agent_template = agent_template or llm.agent_template
-        for transform in self.transforms:
-            agent_template = agent_template.with_additional_post_rendering_transform(transform)
+
+        # Iterate in reverse order to keep the original order of transforms as we are prepending.
+        for transform in self.transforms[::-1]:
+            # Prepend to make sure that user defined transforms are applied before default template transforms.
```
jschweiz, line 327: maybe we should add a info log to tell the users the transformed configured at the agent level (cite their names) will be prepended to the ones from the template (cite their names too), so that in case people see weird behavior, they would link it to how transforms are computed

wayflowcore/tests/transforms/test_summarization_transforms.py
```diff
@@ -1105,3 +1108,50 @@ def test_conversation_summarization_respects_tool_request_response_consistency()
...
+    conv = agent.start_conversation()
+    for m in messages:
+        conv.append_message(m)
+
+    summary = "Summarized conversation"
+    mock_generate_summary = AsyncMock(
+        side_effect=lambda prompt: LlmCompletion(Message(summary), None)
+    )
+
+    with patch_streaming_llm(remote_gemma_llm, "Mock Gemma response") as patched_gemma_llm:
```
jschweiz: couldn't you use the more recent `patch_llm`? This one was made to mock llm calls in a simpler way. I think it also returns the mocked object, so you can assess stuff on it (maybe I just have it locally and never pushed the change to the private repo)

wayflowcore/src/wayflowcore/agent.py
youssef62: Done here [5ea7f19](https://github.com/oracle/wayflow/pull/54/commits/5ea7f19acfb22ea10de61f7f9cefab5923112deb).

wayflowcore/tests/transforms/test_summarization_transforms.py
youssef62: Done here [8c37ed3](https://github.com/oracle/wayflow/pull/54/commits/8c37ed31605ad5a5ad1903552d5ca452f52560df).

nasseur:
Internal regression failed: Build ID #489

nasseur:
Merge Gate failed: Build ID #2132

nasseur:
Internal regression failed: Build ID #549

nasseur:
Merge Gate failed: Build ID #2155

nasseur:
Merge Gate failed: Build ID #2156

nasseur:
Merge Gate succeeded 🍏: Build ID #2157

PR #72, opened 2026-02-04
[fix]: use a single line in the nosec_ignore file per nosec

jschweiz:
Use a single line in the nosec_ignore file per nosec in the codebase


nasseur:
Merge Gate INPROGRESS

nasseur:
Internal regression succeeded 🍏: Build ID #1475

PR #79, opened 2026-02-13
docs-26.1.0-release

younes-makhchan:


26.1.0/_sources/core/changelog.rst.txt
```diff
@@ -0,0 +1,358 @@
+Changelog
+=========
+
+WayFlow |release|
```
dhilloulinoracle: I believe we should remove that part , leave only the 26.1.0 part below?

dhilloulinoracle (approved):

PR #86, opened 2026-02-19
Toolboxes requires_confirmation conversion to Agent Spec

cesarebernardis:
Added conversion of toolboxes requires_confirmation flag in Agent Spec converter.

### DIFF
--- a/wayflowcore/tests/agentspec/test_mcptool.py
+++ b/wayflowcore/tests/agentspec/test_mcptool.py
@@ -104,7 +104,7 @@ def test_mcp_tool_with_requires_confirmation(
     agent = Agent(llm=remotely_hosted_llm, tools=[mcp_toolbox])
     # We need to specify the version to ensure that the requires_confirmation boolean is always dumped
     agentspec_agent = AgentSpecExporter().to_json(
-        agent, agentspec_version=AgentSpecVersionEnum.v26_2_0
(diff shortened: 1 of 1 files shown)


wayflowcore/tests/agentspec/test_mcptool.py
```diff
@@ -86,3 +91,26 @@ def test_mcp_tool_can_be_converted_to_agentspec_and_back(
...
+    requires_confirmation, remotely_hosted_llm, with_mcp_enabled
+):
+    client_transport = StreamableHTTPTransport(url="http://url-to/server")
+    mcp_toolbox = MCPToolBox(
+        client_transport=client_transport, requires_confirmation=requires_confirmation
+    )
+    agent = Agent(llm=remotely_hosted_llm, tools=[mcp_toolbox])
+    # We need to specify the version to ensure that the requires_confirmation boolean is always dumped
+    agentspec_agent = AgentSpecExporter().to_json(
+        agent, agentspec_version=AgentSpecVersionEnum.v26_2_0
```
sonleoracle: Do we just hard-code it like this?

cesarebernardis: I can put `current_version` if you prefer. The point is to make it export the boolean, that in the old version was not there, so it gets removed if the value is False and no export version is specified

PR #87, opened 2026-02-23
Fix MCP session persistence collision in same conversation

cesarebernardis:
The MCP Session persistence mechanism now uses only conversation_id as key. This makes different client transports collide to use the same session if they belong to the same conversation, even though they refer to different MCP servers. We make the session key include also the client transport ID, so that the session is linked to a specific client in a specific conversation.


dhilloulinoracle (approved):

wayflowcore/tests/mcptools/start_mcp_server_alt.py
sonleoracle, line 1: alternative, same tools different names

jschweiz (approved):

nasseur:
Internal regression failed: Build ID #1520

PR #95, opened 2026-02-27
[fix]: no longer ignores tool requests when LLM attempts to both invo…

jschweiz:
Currently, the LLM is not executing tools that are present in a parallel tool calling turn along with a talk_to_user call. We should do the opposite, rather drop the talk_to_user call to force the LLM running all the tool calls before returning to the user

### DIFF
--- a/wayflowcore/src/wayflowcore/executors/_agentexecutor.py
+++ b/wayflowcore/src/wayflowcore/executors/_agentexecutor.py
@@ -347,9 +347,20 @@ class AgentConversationExecutor(ConversationExecutor):
         if len(tool_requests) > 1 and any(
             tr.name in _NON_PARALLEL_TOOL_NAMES for tr in tool_requests
         ):
-            new_message.tool_requests = [
-                tr for tr in new_message.tool_requests if tr.name not in _NON_PARALLEL_TOOL_NAMES
-            ]
+            if all(tr.name in _NON_PARALLEL_TOOL_NAMES for tr in tool_requests):
+                # we have several mnon-parallel tool calls. We'll use the first and drop the rest
+                new_message.tool_requests = new_message.tool_requests[:1]
+            else:
+                logger.warning(
+                    "The LLM tries to call tools and answer to the user. The tools "
+                    "will be run and the LLM will be re-prompted for its answer to "
+                    "the user."
+                )
+                new_message.tool_requests = [
+                    tr
+                    for tr in new_message.tool_requests
+                    if tr.name not in _NON_PARALLEL_TOOL_NAMES
+                ]
 
         state.tool_call_queue.extend(new_message.tool_requests)
         return False
--- a/docs/wayflowcore/source/core/changelog.rst
+++ b/docs/wayflowcore/source/core/changelog.rst
@@ -62,6 +62,8 @@ Possibly Breaking Changes
 Bug fixes
 ^^^^^^^^^
 
+* Fix: conversations no longer replay already-materialized tool results when resuming manager-worker executions, which could previously duplicate internal `send_message` results and break continued execution after serialization.
+
 WayFlow 26.1.1
 --------------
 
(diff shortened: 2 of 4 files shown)


wayflowcore/src/wayflowcore/executors/_agentexecutor.py
```diff
@@ -335,15 +335,16 @@ async def _decide_next_action(
...
-        )
-        if talk_to_user_request:
-            logger.debug(
-                "Parallel tool calling is disabled when using %s tool", _TALK_TO_USER_TOOL_NAME
-            )
-            new_message.tool_requests = [talk_to_user_request]
+        tool_requests = new_message.tool_requests
+
+        if len(tool_requests) > 1 and any(
+            tr.name == _TALK_TO_USER_TOOL_NAME for tr in tool_requests
```
dhilloulinoracle: should we check if it is among a set of "special tools", where for now special tools is mainly the one to talk to users?
Please check if there are other special tools that should go there.
E.g. swarm hand-off tool

wayflowcore/src/wayflowcore/executors/_agentexecutor.py
```diff
@@ -335,15 +335,16 @@ async def _decide_next_action(
...
+        ):
+            # cannot have both talk_to_user tool and other tools. In this case, we
+            # drop the talk to user tool
+            new_message.tool_requests = [
+                tr for tr in new_message.tool_requests if tr.name != _TALK_TO_USER_TOOL_NAME
```
dhilloulinoracle: please filter out if among "special tools"

wayflowcore/src/wayflowcore/executors/_agentexecutor.py
```diff
@@ -335,15 +335,16 @@ async def _decide_next_action(
...
-                "Parallel tool calling is disabled when using %s tool", _TALK_TO_USER_TOOL_NAME
-            )
-            new_message.tool_requests = [talk_to_user_request]
+        tool_requests = new_message.tool_requests
+
+        if len(tool_requests) > 1 and any(
+            tr.name == _TALK_TO_USER_TOOL_NAME for tr in tool_requests
+        ):
+            # cannot have both talk_to_user tool and other tools. In this case, we
+            # drop the talk to user tool
```
paul-cayet: Put a message saying something like "otherwise the model may notify the user that it is waiting for the tool to execute, which is an undesirable behaviour for the user", or something similar

wayflowcore/src/wayflowcore/executors/_agentexecutor.py
```diff
@@ -335,15 +335,16 @@ async def _decide_next_action(
...
+            new_message.tool_requests = [
```
sonleoracle: If we drop talk_to_user tool request in this turn, what about the conversation history in the next turn? Is the talk_to_user tool call completely purged from the chat history? I just want to ensure the chat history is consistent across LLM invocations

wayflowcore/src/wayflowcore/executors/_agentexecutor.py
jschweiz: yes, mainly just thisone, the other we're fine with allowing parallel tool calling

wayflowcore/src/wayflowcore/executors/_agentexecutor.py
jschweiz: added

wayflowcore/src/wayflowcore/executors/_agentexecutor.py
jschweiz: next turn the conversation history doesn't contain the talk_to_user tool anymore, so it's fine

wayflowcore/src/wayflowcore/executors/_agentexecutor.py
jschweiz: added

nasseur:
Internal regression failed: Build ID #696

dhilloulinoracle:
Internal regression failed: Build ID #1541

PR #99, opened 2026-03-01
Normalize property.py default enum casing

### LINKED ISSUE #98: [BUG] - property.py rejects valid enum strings
Discovered bug while utilising with SQLcl MCP... I will contribue a fix.

The bug here is exposed due to a minor case-typo bug in SQLcl (I'll raise with SQLcl team) which makes it impossible to load the SQLcl MCP tools at all. 

The `property.py` __post_init__ doesn't do case-insensitive normalization ("all" → "ALL") before the validation check, so when SQLcl's list_connections tool declares default: "all" against enum: ["OCI", "NON_OCI", "ALL"], __post_init__ treats "all" as an invalid value and raises a ValueError — even though the intent is clearly "ALL". 

The error message is also misleading because is_value_of_expected_type() conflates enum membership and type checking into a single boolean.  It reports "Default value all is not of type StringProperty" rather than telling you the default isn't a member of the enum.

## Steps to Reproduce
1. Define an enum such as {"ALL", "SOME"} in a `Property` definition.
2. Set the `default_value` to "all" (lowercase) while keeping `_validate_default_type=True`.
3. Instantiate the `Property`.

### Expected Result
- The property accepts the default because "all" semantically matches "ALL".

### Actual Result
- `Property.__post_init__` raises a validation error stating the default is not in the enum.

## Impact
- Configuration files that rely on lowercase defaults are rejected.
- Prevents loading previously valid specs, causing feature toggles to break at startup.

gotsysdba:
Fixes #98

Normalize string defaults to the exact enum token before validation, e.g. `"all"` → `"ALL"` and reuse existing enum listings to map case-insensitively without altering validation logic.

Fallback behavior remains unchanged when no case-insensitive match exists.

### DIFF
--- a/wayflowcore/src/wayflowcore/property.py
+++ b/wayflowcore/src/wayflowcore/property.py
@@ -6,6 +6,7 @@
 
 import json
 import logging
+import warnings
 from abc import ABC, abstractmethod
 from copy import deepcopy
 from dataclasses import dataclass, field
@@ -148,6 +149,13 @@ class Property(SerializableObject, ABC):
             lower = self.default_value.lower()
             for ev in self.enum:
                 if isinstance(ev, str) and ev.lower() == lower:
+                    warnings.warn(
+                        f"Default value '{self.default_value}' does not match enum casing; "
+                        f"matched '{ev}' using case-insensitive comparison. "
+                        f"Please update the default to '{ev}'.",
+                        UserWarning,
+                        stacklevel=2,
+                    )
                     object.__setattr__(self, 'default_value', ev)
                     break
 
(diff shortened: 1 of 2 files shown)


wayflowcore/src/wayflowcore/property.py
```diff
@@ -138,6 +138,19 @@ def _validate_or_return_default_value(self, value: Any) -> Any:
...
+        if (
+            self.has_default
+            and self.enum is not None
+            and isinstance(self.default_value, str)
+            and self.default_value not in self.enum
+        ):
+            lower = self.default_value.lower()
+            for ev in self.enum:
+                if isinstance(ev, str) and ev.lower() == lower:
+                    object.__setattr__(self, 'default_value', ev)
```
cesarebernardis, line 159: Please also raise a UserWarning to inform the user about this lower-case matching, and ask her to change the default

gotsysdba: UserWarning raised.

PR #102, opened 2026-03-02
Fix: Use OciSessionAuth for OCI security token authentication.

qiuosier:
The current implementation of using `HttpxOciAuth` is causing an error as it is an abstract class.
```
TypeError: Can't instantiate abstract class HttpxOciAuth without an implementation for abstract method '_refresh_signer'
```

This PR fixes the issue by using the `OciSessionAuth` class provided by `oci_openai`, as documented by the example [here](https://github.com/oracle-samples/oci-openai?tab=readme-ov-file#using-the-oci-openai-synchronous-client).


dhilloulinoracle:
I guess due to a change in the oci-openai package?
 @sonleoracle if you could confirm and indicate what version triggers it

nasseur:
Merge Gate failed: Build ID #2249

nasseur:
Merge Gate failed: Build ID #2250

PR #110, opened 2026-03-06
feat OpenAICompatibleEmbeddingModel supports optional api_key

sonleoracle:
Add an optional api_key field to OpenAICompatibleEmbeddingModel that falls back to OPENAI_API_KEY in the env

try it out with:

```python
from wayflowcore.embeddingmodels import OpenAICompatibleEmbeddingModel

model = OpenAICompatibleEmbeddingModel(
    model_id="text-embedding-3-small",
    base_url="https://api.openai.com",  # uses /v1/embeddings automatically
    api_key=os.environ["OPENAI_API_KEY"],
)

emb = model.embed(["ping"])[0]
print(len(emb))
```

### DIFF
--- a/wayflowcore/tests/test_embeddingmodels.py
+++ b/wayflowcore/tests/test_embeddingmodels.py
@@ -472,6 +472,14 @@ def test_invalid_open_api_key():
             OpenAIEmbeddingModel,
             {"model_id": "text-embedding-3-small", "api_key": "fake-api-key"},
         ),
+        (
+            OpenAICompatibleEmbeddingModel,
+            {
+                "model_id": "text-embedding-3-small",
+                "base_url": "https://api.openai.com",
+                "api_key": "fake-api-key",
+            },
+        ),
     ],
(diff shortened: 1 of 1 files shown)


nasseur:
Internal regression failed: Build ID #724

wayflowcore/tests/test_embeddingmodels.py
```diff
@@ -331,6 +334,60 @@ def test_embedding_model_multiple_embeds_vllm_ollama(
     assert all(len(embedding) == embedding_dim for embedding in embeddings)
 
 
+def test_openai_compatible_embedding_model_api_key_sets_auth_header(
```
cesarebernardis, line 337: Can you add a test to make sure that the new api_key parameter does not get serialized? It should be already green, but it's mainly to ensure that we do not dump it by mistake at some point

nasseur:
Internal regression succeeded 🍏: Build ID #732

PR #133, opened 2026-04-10
fix: align Agent Spec tracing and transform serialization compatibility

paul-cayet:


dhilloulinoracle:
Internal regression failed: Build ID #789

PR #152, opened 2026-05-06
Improve default requirements in remote Components

cesarebernardis:
This PR refines URL templating behavior in ApiCallStep and RemoteTool. It now requires url_allow_list only when placeholders are used in the destination part of a URL (scheme, host, or port), adds a runtime warning for localhost/private IP literal destinations, and updates tests, docs, and examples to reflect the recommended patterns.

PR #156, opened 2026-05-07
Add customizable AgentSpec component loading policies

ljupche98:


wayflowcore/src/wayflowcore/agentspec/runtimeloader.py
```diff
@@ -18,6 +19,10 @@
 
 FieldName: TypeAlias = str
 RuntimeComponentsRegistryT: TypeAlias = Mapping[str, Union[RuntimeComponent, Any]]
+_DEFAULT_BLOCKED_COMPONENTS: tuple[str, ...] = (
```
cesarebernardis: We should make it extend the basic one from pyagentspec. If we will add something there, we will miss it here.

wayflowcore/src/wayflowcore/agentspec/runtimeloader.py
```diff
@@ -18,6 +19,10 @@
...
+    "StdioTransport",
+    "PluginStdioTransport",
+)
 
 
 class AgentSpecLoader:
```
cesarebernardis, line 36: Arguably we could give a try at making this extend the adapter agnostic loader of pyagentspec, can you give it a quick try and see if it works without major modifications? We will likely be able to remove a lot of logic from here. It's not necessarily the scope of this PR though, if problems surge we can do it another time.

cesarebernardis (approved):

ljupche98:
These changes will be included in PR #152

PR #159, opened 2026-05-11
Checkpointing APIs

sonleoracle:
# Summary

This PR adds conversation checkpointing to WayFlowCore.

Checkpointing is integrated into normal conversation execution: conversations can be started with a `Checkpointer`, execution can save checkpoints at configured boundaries, and later runs can restore and resume from a saved checkpoint.

# What Changed

- Added the `wayflowcore.checkpointing` module with:
  - `ConversationCheckpoint`
  - `CheckpointingInterval`
  - `Checkpointer`
  - `InMemoryCheckpointer`
  - `PostgresCheckpointer`
  - `OracleDatabaseCheckpointer`

- Added checkpoint-aware `start_conversation(...)` support for conversations, flows, agents, swarms, manager workers, A2A agents, and OpenAI Responses service flows.

- Added a checkpoint execution listener that saves checkpoints after:
  - conversation turns
  - LLM turns
  - all internal turn boundaries

# Example

```python
from wayflowcore.checkpointing import InMemoryCheckpointer

checkpointer = InMemoryCheckpointer()

conversation = flow.start_conversation(
    messages="Plan a trip to Bern",
    checkpointer=checkpointer,
)

conversation.execute()

checkpoint = checkpointer.load_latest(conversation.conversation_id)

restored = flow.start_conversation(
    conversation_id=conversation.conversation_id,
    checkpoint_id=checkpoint.checkpoint_id,
    checkpointer=checkpointer,
)

restored.execute()
```


dhilloulinoracle:
Internal regression failed: Build ID #857

dhilloulinoracle:
Internal regression failed: Build ID #858

dhilloulinoracle:
Internal regression failed: Build ID #859

dhilloulinoracle:
Internal regression failed: Build ID #860

dhilloulinoracle:
Internal regression failed: Build ID #861

dhilloulinoracle:
Internal regression failed: Build ID #862

dhilloulinoracle:
Internal regression failed: Build ID #863

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #864

dhilloulinoracle:
Internal regression failed: Build ID #865

dhilloulinoracle:
Internal regression failed: Build ID #867

dhilloulinoracle:
Internal regression failed: Build ID #868

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #871

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #872

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #873

dhilloulinoracle:
Internal regression failed: Build ID #875

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #877

dhilloulinoracle:
Internal regression failed: Build ID #884

dhilloulinoracle:
Internal regression failed: Build ID #886

dhilloulinoracle:
Internal regression failed: Build ID #889

dhilloulinoracle:
Internal regression failed: Build ID #890

dhilloulinoracle:
Internal regression failed: Build ID #923

dhilloulinoracle:
Internal regression failed: Build ID #926

dhilloulinoracle:
Internal regression failed: Build ID #927

dhilloulinoracle:
Internal regression failed: Build ID #928

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #929

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #930

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #931

dhilloulinoracle:
Internal regression failed: Build ID #932

dhilloulinoracle:
Internal regression failed: Build ID #935

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #936

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #937

dhilloulinoracle:
Internal regression failed: Build ID #938

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #939

dhilloulinoracle:
Internal regression failed: Build ID #940

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #941

dhilloulinoracle:
Internal regression failed: Build ID #942

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #943

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #944

wayflowcore/src/wayflowcore/conversation.py
```diff
@@ -85,21 +95,32 @@ def _register_conversation(conversation: "Conversation") -> Generator[None, Any,
...
-    component: ConversationalComponent
+    component: "ConversationalComponent"
     state: "ConversationExecutionState"
     inputs: Dict[str, Any]
     message_list: MessageList
     status: Optional[ExecutionStatus]
     token_usage: TokenUsage = field(default_factory=TokenUsage, init=False)
-    conversation_id: str = ""  # deprecated
-
+    conversation_id: str = ""
```
cesarebernardis, line 104: Conversation has `id` and `conversation_id`, and they are not the same thing? Please choose a better name.

wayflowcore/src/wayflowcore/agent.py
```diff
@@ -398,7 +403,16 @@ def start_conversation(
...
             Message list to which the agent will participate
         conversation_id:
-            Conversation id of the parent conversation.
+            Durable conversation id used for resume, storage, and usage accounting.
+        checkpointer:
+            Optional checkpoint backend used to restore and persist this conversation.
+        checkpoint_id:
+            Optional checkpoint identifier to restore. Requires both ``checkpointer`` and
+            ``conversation_id``.
+        _runtime_conversation_id:
```
cesarebernardis: A private parameter should probably not be in the public docstring. Same in the other subclasses. 

It's also unclear why we need `conversation_id` and `_runtime_conversation_id`. What is a "runtime conversation"? When should I use one or the other? What if they are different?

After reviewing more of this PR it seems to me that one is the id of the root conversation, the other is the id of the actual conversation being created, but it's impossible to get this without reading through the whole code, and I am not even sure how I am supposed to use them differently. At least we need to find a better name without breaking backward compatibility. An alternative would be to have a private method with better names that we use internally where conversation_id and _runtime_conversation_id can actually be different (e.g., _start_subconversation or so), since `_runtime_conversation_id` is something that should not be used publicly, as it's a private argument.

wayflowcore/src/wayflowcore/conversationalcomponent.py
```diff
@@ -66,8 +81,19 @@ def start_conversation(
...
+        checkpointer: Optional["Checkpointer"] = None,
+        checkpoint_id: Optional[str] = None,
+        _runtime_conversation_id: Optional[str] = None,
+        _attach_checkpointer: bool = True,
     ) -> "Conversation":
-        pass
+        """Start a fresh conversation or restore one from checkpoint storage.
+
+        ``conversation_id`` is the durable conversation identifier used for resume and
+        storage. For fresh conversations, ``_runtime_conversation_id`` controls the
```
cesarebernardis: We should not mention a private parameter in a public docstring

wayflowcore/src/wayflowcore/agentserver/_storagehelpers.py
```diff
@@ -4,106 +4,25 @@
...
-from wayflowcore.serialization import autodeserialize
-from wayflowcore.serialization.context import DeserializationContext
-from wayflowcore.tools import Tool
+from wayflowcore.datastore.oracle import OracleDatabaseConnectionConfig
+from wayflowcore.datastore.postgres import PostgresDatabaseConnectionConfig
 
 logger = logging.getLogger(__name__)
 
 
 def _prepare_postgres_datastore(
```
cesarebernardis: These methods seem a bit useless, can use the originals directly

wayflowcore/src/wayflowcore/agentserver/serverstorageconfig.py
```diff
@@ -4,52 +4,11 @@
...
-from typing import Dict, Optional
 
-from wayflowcore.datastore import Datastore, Entity
-from wayflowcore.property import IntegerProperty, StringProperty
+from wayflowcore.checkpointing import StorageConfig
 
 
 @dataclass
-class ServerStorageConfig:
+class ServerStorageConfig(StorageConfig):
```
cesarebernardis, line 13: Also here, do we really need this class?

wayflowcore/tests/integration/test_checkpointing.py
```diff
@@ -0,0 +1,528 @@
...
+from wayflowcore.managerworkers import ManagerWorkers
+from wayflowcore.steps import FlowExecutionStep, OutputMessageStep
+from wayflowcore.swarm import Swarm
+
+from ..serialization.test_assistant_serialization import create_flow
+from ..testhelpers.dummy import DoNothingStep, DummyModel
+
+
+def _build_nested_flow(
+    *,
```
cesarebernardis: Nit: cleanup these * put in arbitrary places

wayflowcore/tests/integration/test_checkpointing.py
```diff
@@ -0,0 +1,528 @@
...
+        (_build_checkpointable_swarm, "swarm-checkpoint"),
+        (_build_checkpointable_managerworkers, "managerworkers-checkpoint"),
+    ],
+)
+def test_multi_agent_checkpointing_supports_resume_and_time_travel(
+    builder,
+    conversation_id: str,
+) -> None:
+    checkpointer = InMemoryCheckpointer()
+    component, llm = builder()
```
cesarebernardis: Nit:
- a bit weird that _build_checkpointable_agent_component and others return also the LLM, while the name does not state that
- arguably those methods are used here only and the outputs are the same, so probably you don't really need to return the LLM separately and output could be set a priori

wayflowcore/tests/integration/test_checkpointing.py
```diff
@@ -0,0 +1,528 @@
...
+        group_manager=manager_agent,
+        workers=[worker_agent],
+        name="checkpoint_managerworkers",
+        id="checkpoint_managerworkers",
+    )
+    return managerworkers, manager_llm
+
+
+def test_flow_checkpoint_restore_rejects_generated_step_name_reinstantiation() -> None:
+    checkpointer = InMemoryCheckpointer()
```
cesarebernardis: I don't see any test with different checkpointers

wayflowcore/tests/integration/test_checkpointing.py
```diff
@@ -0,0 +1,528 @@
...
+    conversation = original_swarm.start_conversation(
+        conversation_id="swarm-generated-agent-name-restart",
+        checkpointer=checkpointer,
+    )
+    original_thread = conversation.state.agents_and_threads[original_first_agent.name][
+        original_second_agent.name
+    ]
+    conversation.state._create_subconversation_for_thread(original_thread)
+    conversation.state.current_thread = original_thread
+    _save_live_conversation_checkpoint(checkpointer, conversation)
```
cesarebernardis: Not a fan of using these many private apis in tests, can't we have a clean flow that users might actually encounter?

wayflowcore/tests/integration/test_checkpointing.py
```diff
@@ -0,0 +1,528 @@
...
+    restarted_flow, _ = build_flow()
+
+    with pytest.raises(ValueError, match="stable step names"):
+        restarted_flow.start_conversation(
+            conversation_id=conversation.conversation_id,
+            checkpointer=checkpointer,
+        )
+
+
+def test_flow_checkpoint_restore_relinks_nested_flow_parent_for_interrupt_inheritance() -> None:
```
cesarebernardis: I think it would be good to have similar tests where the component is serialized and deserialized (and its conversation should resume nicely on the deserialized version)

wayflowcore/src/wayflowcore/checkpointing/checkpoint_state.py
```diff
@@ -0,0 +1,251 @@
...
+    from wayflowcore.executors._flowexecutor import FlowConversationExecutionState
+    from wayflowcore.executors._managerworkersconversation import (
+        ManagerWorkersConversationExecutionState,
+    )
+    from wayflowcore.executors._swarmconversation import SwarmConversationExecutionState
+
+
+_COMPONENT_ID_ERROR_HINT = "Restart-safe checkpoint restore requires stable component ids."
+_STEP_NAME_ERROR_HINT = "Restart-safe checkpoint restore requires stable step names."
+_AGENT_NAME_ERROR_HINT = "Restart-safe checkpoint restore requires stable agent names."
```
cesarebernardis: Why are these split from the warning? They don't seem to be reused, are they?

wayflowcore/src/wayflowcore/checkpointing/checkpoint_state.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+class _CheckpointRestoreCompatibilityError(ValueError):
```
cesarebernardis: Not a big fan of having this as a private class, unless we catch every exception of this kind that might occur through a public API (and that's not the case currently). Either make it public or don't create it at all.

wayflowcore/src/wayflowcore/checkpointing/checkpoint_state.py
```diff
@@ -0,0 +1,251 @@
...
+    conversation: "Conversation",
+    checkpoint_id: Optional[str] = None,
+    component_id: Optional[str] = None,
+    metadata: Optional[Dict[str, Any]] = None,
+) -> "ConversationCheckpoint":
+    from wayflowcore.checkpointing.checkpointer import ConversationCheckpoint
+
+    if not _supports_checkpointing(conversation.component):
+        raise NotImplementedError(
+            "Checkpointing conversations that contain `OciAgent` is not supported yet."
```
cesarebernardis: The error line and the check don't match, or better, they make assumptions that might not hold reading the code: `_supports_checkpointing` has nothing to do with `OciAgent` (unless you know exactly the internals of `_supports_checkpointing`). Either:
- raise the error as part of the method, if you want custom exception texts
- raise a generic error here without mentioning `OciAgent` directly

wayflowcore/src/wayflowcore/checkpointing/checkpoint_state.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+# Checkpoint eligibility
+
+
+def _supports_checkpointing(component: "ConversationalComponent") -> bool:
+    from wayflowcore.ociagent import OciAgent
+
+    return not any(
+        isinstance(nested_component, OciAgent)
```
cesarebernardis: Nit: I would rather that the component itself knows if it supports checkpointing or not, instead of having a third party check, but it's not a big concern, it's acceptable this way too

wayflowcore/src/wayflowcore/steps/parallelflowexecutionstep.py
```diff
@@ -255,9 +255,9 @@ async def _invoke_step_async(
                     if input_name in set(p.name for p in flow.input_descriptors)
                     # We extract the inputs needed by this specific flow
                 },
-                sub_conversation_id=flow.id,
+                sub_conversation_id=f"parallel_subflow_{index}",
```
cesarebernardis: Why this change? 

If it's because originally if the flow was reused inside this step and somewhere else conversations could conflict, I think we still have the same problem now, if the flow is reused in multiple steps of this type. You might have to include the step id. If it's for another reason, please explain and leave a comment

wayflowcore/src/wayflowcore/serialization/context.py
```diff
@@ -38,6 +38,42 @@ def _create_component_type_to_plugin_mapping(
     return component_types_to_plugins
 
 
+class MissingDeserializationReferenceError(ValueError):
+    """Raised when deserialization encounters a reference missing from the root object."""
+
+
+def _iter_nested_components(value: Any) -> List["Component"]:
```
cesarebernardis: Nit: `iter...` makes me think that this is an iterator, but it just returns a list. Maybe let's just call it `_get...`

wayflowcore/src/wayflowcore/executors/_managerworkersconversation.py
```diff
@@ -26,11 +26,15 @@
...
     subconversations: Dict[str, Union["AgentConversation", "ManagerWorkersConversation"]]
+    conversation_id: str = ""
 
     def _create_subconversation_for_agent(
         self, agent: Union[Agent, ManagerWorkers]
     ) -> Union["AgentConversation", "ManagerWorkersConversation"]:
-        subconv = agent.start_conversation()
+        subconv = agent.start_conversation(
+            conversation_id=self.conversation_id or None,
+            _runtime_conversation_id=agent.name,
```
cesarebernardis: Using a name as id sounds a bit dangerous. Why not agent's id?

wayflowcore/src/wayflowcore/executors/_flowexecutor.py
```diff
@@ -246,7 +246,7 @@ class FlowConversationExecutor(ConversationExecutor):
 
     @staticmethod
     def make_key_for_step(assistant_step: Step, key: str) -> str:
-        return str(assistant_step.id) + "_" + key
+        return str(assistant_step.name) + "_" + key
```
cesarebernardis: Why changing to name? I guess wayflow's steps must have unique names too, so it should be fine, but why are we changing it?

wayflowcore/src/wayflowcore/executors/_agentconversation.py
```diff
@@ -65,15 +65,22 @@ def _get_all_sub_conversations(self) -> List["Conversation"]:
             sub_conversations += [self.state.current_flow_conversation]
         return sub_conversations
 
+    @staticmethod
+    def _sub_component_conversation_key(component: ConversationalComponent) -> str:
+        """Return the canonical runtime key for a subcomponent conversation slot."""
+        return f"{component.__class__.__name__}:{component.name}"
```
cesarebernardis: Why the name as key, it might not be unique

wayflowcore/src/wayflowcore/checkpointing/checkpointeventlistener.py
```diff
@@ -0,0 +1,205 @@
...
+    """Build listener metadata describing why this checkpoint was saved."""
+    checkpoint_metadata: Dict[str, Any] = {
+        "save_reason": save_reason,
+        "current_step_name": conversation.current_step_name,
+        "message_count": len(conversation.message_list.messages),
+    }
+    if conversation.status is not None:
+        checkpoint_metadata["status_type"] = type(conversation.status).__name__
+    if isinstance(event, AgentExecutionIterationStartedEvent):
+        checkpoint_metadata["event_type"] = AgentExecutionIterationStartedEvent.__name__
```
cesarebernardis: `event_type` is set only with these two events? Can't it just be `checkpoint_metadata["event_type"] = event.__class__.__name__`

wayflowcore/src/wayflowcore/checkpointing/checkpointeventlistener.py
```diff
@@ -0,0 +1,205 @@
...
+
+        self._latest_checkpoint_boundary_event = event
+
+        checkpointed_conversation = _find_checkpointed_conversation(
+            self.conversation, event.execution_state
+        )
+        if checkpointed_conversation is None:
+            return
+
+        match self.checkpointer.checkpointing_interval:
```
cesarebernardis, line 129: Are these different intervals tested with (at least) both flows and agents?

wayflowcore/src/wayflowcore/checkpointing/datastorecheckpointer.py
```diff
@@ -0,0 +1,512 @@
...
+                "This can happen if multiple writers checkpoint the same conversation "
+                "concurrently or if the checkpoint datastore was modified manually. Use a "
+                "single writer per conversation or load a specific checkpoint by checkpoint_id."
+            )
+        checkpoints = self._sort_checkpoints(
+            [self._entity_to_checkpoint(entity) for entity in entities]
+        )
+        return checkpoints[-1]
+
+    def load(self, conversation_id: str, checkpoint_id: str) -> ConversationCheckpoint:
```
cesarebernardis, line 242: Nit: personal preference, but when I see `load` and `save` on a class, I expect them to do it on themselves, not another object. I would probably prefer `load_checkpoint` and `save_checkpoint`, but not a big deal.

wayflowcore/src/wayflowcore/checkpointing/datastorecheckpointer.py
```diff
@@ -0,0 +1,512 @@
...
+    """Checkpointer backed by an in-memory datastore."""
+
+    def __init__(
+        self,
+        storage_config: Optional[StorageConfig] = None,
+        checkpointing_interval: CheckpointingInterval = CheckpointingInterval.CONVERSATION_TURNS,
+    ) -> None:
+        resolved_storage_config = storage_config or StorageConfig()
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore", message=f"{_INMEMORY_USER_WARNING}*")
```
cesarebernardis: Doesn't the warning apply also in this case? I would let it go through, or maybe put one more suited for the checkpointer

docs/wayflowcore/source/core/code_examples/howto_checkpointing.py
```diff
@@ -0,0 +1,71 @@
...
+# .. end-##_Resume_the_latest_checkpoint
+
+# .. start-##_Load_a_specific_checkpoint
+# Checkpoints are ordered oldest -> newest.
+checkpoints = checkpointer.list_checkpoints(conversation_id)
+
+previous_checkpoint = checkpoints[-2]
+rewound_conversation = agent.start_conversation(
+    conversation_id=conversation_id,
+    checkpoint_id=previous_checkpoint.checkpoint_id,
```
sonleoracle, line 57: Would it be better for `start_conversation` to take a `checkpoint` object rather than a `checkpoint_id` ?

wayflowcore/src/wayflowcore/agentserver/_storagehelpers.py
sonleoracle: Removed them

wayflowcore/tests/integration/test_checkpointing.py
sonleoracle: cleaned up

wayflowcore/tests/integration/test_checkpointing.py
sonleoracle: cleaned up

wayflowcore/tests/integration/test_checkpointing.py
sonleoracle: added

wayflowcore/tests/integration/test_checkpointing.py
sonleoracle: Yes there are parametrized tests in test_checkpointing.py

wayflowcore/src/wayflowcore/serialization/context.py
sonleoracle: changed to _get_nested_components

wayflowcore/src/wayflowcore/checkpointing/checkpointeventlistener.py
sonleoracle: it's now just the direct assignment, yes it's only those two events

wayflowcore/src/wayflowcore/checkpointing/checkpoint_state.py
sonleoracle: now the conversational component knows if it supports checkpointing or not. composite components like manager workers also check its nested components for this too.

wayflowcore/src/wayflowcore/checkpointing/checkpoint_state.py
sonleoracle: made the error message generic

wayflowcore/src/wayflowcore/checkpointing/checkpoint_state.py
sonleoracle: I removed this file entirely (and inlined these)

wayflowcore/src/wayflowcore/steps/parallelflowexecutionstep.py
sonleoracle: yes as you said, it's an existing problem, but it's out of scope of this PR, so I removed it from this PR

wayflowcore/src/wayflowcore/checkpointing/checkpointeventlistener.py
sonleoracle: yes there are now parametrized tests in test_checkpointing.py

wayflowcore/src/wayflowcore/checkpointing/datastorecheckpointer.py
sonleoracle: the warning goes through now

wayflowcore/src/wayflowcore/conversation.py
sonleoracle: `id` is an instance-specific ID, while `conversation_id` identifies the whole conversation thread, from the top-level conversation to all child subconversations, e.g. flow, manager workers, swarms. `conversation_id` is an existing parameter, this PR just repurposes it into the thread ID, so I don't think the name needs to be changed. I refactored the code to make more explicit what id is used for what.

wayflowcore/src/wayflowcore/agentserver/serverstorageconfig.py
sonleoracle: this class is now an alias to the new `StorageConfig`. I would prefer not to remove it since it has been documented in our docs, so let's keep it for now to maintain backward compatibility

wayflowcore/src/wayflowcore/checkpointing/datastorecheckpointer.py
sonleoracle: `load` and `save` are agreed APIs.

wayflowcore/src/wayflowcore/checkpointing/checkpoint_state.py
sonleoracle: I made it public, it's useful so that users would know if the checkpoint can be restored or not. The usual cause of this error is that the checkpoint was saved by a component whose component ids are different from the current one that's trying to load the checkpoint.

wayflowcore/src/wayflowcore/conversationalcomponent.py
sonleoracle: I removed `_runtime_conversation_id` and added `_start_conversation_impl` which has `parent_conversation` to indicate the whole conversation thread

wayflowcore/tests/integration/test_checkpointing.py
sonleoracle: No tests are using private APIs anymore

wayflowcore/src/wayflowcore/executors/_managerworkersconversation.py
sonleoracle: In main branch, swarm and manager worker is using names as unique identifiers, since they can surface in prompts sent to LLMs. In this PR though, I refactored manager workers and swarms to use agent.id as the identifier in routing and orchestration so that it's the same as Flow and Agent internals. The prompts sent to LLMs still contain names.

wayflowcore/src/wayflowcore/executors/_agentconversation.py
sonleoracle: changed it back to id-based instead of names

wayflowcore/src/wayflowcore/executors/_flowexecutor.py
sonleoracle: It was probably because manager workers and swarms were using names as identifiers. As this id->name change is not necessary, I changed it back to how it was before, i.e. `return str(assistant_step.id) + "_" + key`

wayflowcore/src/wayflowcore/agent.py
sonleoracle: You're right. I removed the concept of `_runtime_conversation_id`, now it's represented by an internal private `_start_conversation_impl` method that has a `parent_conversation` parameter (so that, e .g., child conversations can reuse the parent's `conversation_id`, so now it should be clearer that conversation_id refers to the thread identifier (persisted when checkpointed).

dhilloulinoracle:
Internal regression failed: Build ID #972

dhilloulinoracle:
Internal regression failed: Build ID #973

dhilloulinoracle:
Internal regression failed: Build ID #974

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #975

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #981

dhilloulinoracle:
Internal regression failed: Build ID #982

PR #162, opened 2026-05-15
  Improve retry policy and test reliability

sonleoracle:
## Summary

  - Move retry classification into provider-agnostic HTTP/transport helpers.
  - Avoid eager imports from optional dependencies during retry handling (like `openai`).
  - Preserve retry behavior for status, timeout, connection, and TLS error shapes.

  ## Test reliability

  - Fix IPv6 localhost port allocation for tests that start local servers, including MCP and A2A coverage.
  - Fix the test retry decorator so it works with failures raised inside `pytest.raises`.
  - Use `sys.executable` for test subprocesses so helper servers run under the correct python interpreter.

### DIFF
--- a/wayflowcore/src/wayflowcore/models/_requesthelpers.py
+++ b/wayflowcore/src/wayflowcore/models/_requesthelpers.py
@@ -325,7 +325,7 @@ def _is_transport_exception_without_response(exc: BaseException) -> bool:
         return False
 
     status_code = _get_http_status_code_from_exception(exc)
-    return status_code is None or status_code <= 0 or status_code == 408 or status_code >= 500
+    return status_code is None or status_code <= 0 or status_code == 408
 
 
 def _classify_http_exception_for_retry(
--- a/wayflowcore/src/wayflowcore/models/_requesthelpers.py
+++ b/wayflowcore/src/wayflowcore/models/_requesthelpers.py
@@ -314,18 +314,44 @@ def _get_http_error_text_from_exception(exc: BaseException) -> str:
     return _stringify_response_error(body)
 
 
-def _is_transport_exception_without_response(exc: BaseException) -> bool:
+_PRE_RESPONSE_TRANSPORT_EXCEPTION_CLASS_NAMES = {
+    "APIConnectionError",
+    "APITimeoutError",
+    "Timeout",
+}
+
+
+def _is_transport_failure_without_http_response(exc: BaseException) -> bool:
+    """Return whether an exception is a network/client failure before an HTTP response."""
+    # Native httpx transport failures do not represent HTTP responses.
     if isinstance(exc, httpx.TransportError):
         return True
 
-    # Some OpenAI-compatible SDKs expose connection/timeout failures as their
-    # own exception classes with a request and sometimes a synthetic status_code,
-    # but no response. Treat only transport-shaped statuses this way.
-    if getattr(exc, "request", None) is None or getattr(exc, "response", None) is not None:
+    # SDK status errors normally carry a response and must be classified by HTTP
+    # status. Only SDK wrapper errors with a request and no response can be
+    # pre-response transport failures.
+    if getattr(exc, "request", None) is None:
+        return False
+    if getattr(exc, "response", None) is not None:
         return False
 
+    # No real HTTP response exists. Treat absent/invalid status values and the
+    # request-timeout status as transport failures.
     status_code = _get_http_status_code_from_exception(exc)
-    return status_code is None or status_code <= 0 or status_code == 408
+    if status_code is None or status_code <= 0 or status_code == 408:
+        return True
+
+    if not 500 <= status_code < 600:
+        return False
+
+    # LiteLLM/OpenAI connection and timeout wrappers can attach synthetic 5xx
+    # statuses despite having no response. Only transport-shaped wrappers get
+    # this treatment; real no-response APIError 5xx exceptions stay on the HTTP
+    # policy path. Avoid importing optional provider SDKs here; their stable
+    # class names are visible through the MRO when those SDKs are installed.
+    return any(
+        cls.__name__ in _PRE_RESPONSE_TRANSPORT_EXCEPTION_CLASS_NAMES for cls in type(exc).__mro__
+    )
 
 
 def _classify_http_exception_for_retry(
@@ -336,7 +362,7 @@ def _classify_http_exception_for_retry(
     # classes. Keep retry classification based on stable HTTP attributes and
     # httpx transport causes so importing those optional SDKs is not required.
     for current in _iter_exception_chain(exc):
-        if _is_transport_exception_without_response(current):
+        if _is_transport_failure_without_http_response(current):
             if _is_tls_or_cert_error(current):
                 return None
             return None, None
--- a/wayflowcore/src/wayflowcore/models/_requesthelpers.py
+++ b/wayflowcore/src/wayflowcore/models/_requesthelpers.py
@@ -23,6 +23,7 @@ from typing import (
     Callable,
     Dict,
     Generator,
+    Iterator,
     Mapping,
     Optional,
     Tuple,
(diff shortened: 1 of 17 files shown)


dhilloulinoracle:
Internal regression failed: Build ID #881

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #882

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #885

dhilloulinoracle:
Internal regression failed: Build ID #887

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #888

dhilloulinoracle:
Internal regression failed: Build ID #891

dhilloulinoracle:
Internal regression failed: Build ID #893

dhilloulinoracle:
Internal regression failed: Build ID #894

dhilloulinoracle:
Internal regression failed: Build ID #895

dhilloulinoracle:
Internal regression failed: Build ID #896

dhilloulinoracle:
Internal regression failed: Build ID #897

dhilloulinoracle:
Internal regression failed: Build ID #898

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #899

wayflowcore/src/wayflowcore/models/_requesthelpers.py
```diff
@@ -255,6 +264,123 @@ def _classify_oci_service_error_for_retry(
...
+    seen: set[int] = set()
+    while pending:
+        current = pending.pop()
+        if id(current) in seen:
+            continue
+        yield current
+        seen.add(id(current))
+        # Provider SDKs sometimes wrap the transport exception in __cause__,
+        # and framework code may use __context__; walk both links when present.
+        if current.__context__ is not None:
```
paul-cayet, line 278: Is __context__ always there? Should we have some hasattr check?

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #901

wayflowcore/src/wayflowcore/models/_requesthelpers.py
sonleoracle: this helper function is walking through exception chains, and exception objects always contain the `__context__` attribute, so it should always be there

wayflowcore/tests/transforms/conftest.py
```diff
@@ -70,6 +72,8 @@ def find_datastore_by_schema(pool, schema):
 
 MESSAGE_SUMMARIZATION_CACHE_COLLECTION_NAME = "messages_cache"
 CONVERSATION_SUMMARIZATION_CACHE_COLLECTION_NAME = "conversations_cache"
+AGENTSPEC_LIVE_LLM_GENERATION_CONFIG = AgentSpecLlmGenerationConfig(max_tokens=512)
+AGENTSPEC_LIVE_LLM_RETRY_POLICY = AgentSpecRetryPolicy(max_attempts=0, request_timeout=30.0)
```
cesarebernardis: Should be fixtures I think

wayflowcore/tests/transforms/conftest.py
```diff
@@ -248,7 +252,11 @@ def converted_wayflow_agent_with_message_summarization_transform_from_agentspec(
         },
     )
     summarization_llm_config = AgentSpecVllmConfig(
```
cesarebernardis: Arguably this could be a fixture as well, reused below

wayflowcore/tests/test_swarm.py
```diff
@@ -1355,17 +1355,17 @@ def _setup_swarm_for_multiple_tool_calling(vllm_responses_llm, raise_exceptions)
...
     vllm_responses_llm,
 ):
     """
-    Failure rate:          0 out of 20
-    Observed on:           2026-01-28
-    Average success time:  5.19 seconds per successful attempt
-    Average failure time:  No time measurement
-    Max attempt:           3
-    Justification:         (0.05 ** 3) ~= 9.4 / 100'000
+    Failure rate:          4 out of 20
```
cesarebernardis, line 1363: This is a huge performance drop, I think we need to investigate why this is happening. I know that it's likely a model issue, but I think we need to make sure that it is the case. Not in the scope of this PR though.

cesarebernardis (approved):

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #905

wayflowcore/tests/test_swarm.py
sonleoracle: Yes I am actively investigating this issue and will address it in a follow-up PR.

wayflowcore/tests/transforms/conftest.py
sonleoracle: turned this into a fixture and inlined the generation config and retry policy

wayflowcore/tests/transforms/conftest.py
sonleoracle: addressed in comment https://github.com/oracle/wayflow/pull/162#discussion_r3272312351

wayflowcore/tests/models/test_ocigenaimodel.py
```diff
@@ -413,15 +414,22 @@ def test_oci_cohere_invalid_content_and_unsupported_types(messages, match):
 
 
 def test_oci_model_raises_warning_when_parameters_are_not_supported(grok_oci_llm, caplog):
+    # The shared live-model retry policy disables HTTP retries. This test needs one
+    # retry so Wayflow can handle OCI's unsupported-parameter error by removing
+    # `stop` from the generation config and resubmitting the request.
+    grok_oci_llm.retry_policy = RetryPolicy(max_attempts=1, request_timeout=30.0)
```
paul-cayet, line 420: This is not scoped to this test is it? Do we need to store previous value of this attribute and restore it upon completion? What about when tests are run concurrently?

wayflowcore/tests/testhelpers/test_server_port_retrieval.py
```diff
@@ -24,3 +26,38 @@ def test_get_available_port_returns_unique_ports_in_parallel(
         futures = [executor.submit(get_available_port, str(temp_dir)) for _ in range(request_count)]
     ports = {future.result() for future in futures}
     assert len(ports) == request_count
+
+
+def test_get_available_port_skips_ports_occupied_on_ipv6(
+    monkeypatch, tmp_path_factory: pytest.TempPathFactory
+):
+    if not socket.has_ipv6:
+        pytest.skip("IPv6 is not available on this host")
```
paul-cayet, line 35: skip or raise? Same below

wayflowcore/tests/conftest.py
```diff
@@ -195,6 +202,7 @@ def oci_user_authentication_config():
     "host_port": llama_api_url,
     "model_id": "meta-llama/Meta-Llama-3.1-8B-Instruct",
     "generation_config": {"max_tokens": 512},
+    "retry_policy": _create_live_llm_test_retry_policy(),
```
paul-cayet, line 205: Why not a fixture?

paul-cayet (approved):

wayflowcore/tests/conftest.py
sonleoracle: Because these LLM config dicts are defined at a module level. If we want this to be a fixture, then all of these LLM config dicts need to become fixtures too. For now, just adding a helper like this is better I think.

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #906

wayflowcore/tests/models/test_ocigenaimodel.py
sonleoracle: the `grok_oci_llm` is a function-scoped fixture, and thus mutating it is specific to this test and this run, so it's safe even when running tests concurrently

wayflowcore/tests/testhelpers/test_server_port_retrieval.py
sonleoracle: it should be skipped because the host machine may not have ipv6. this test is only meaningful on hosts with ipv6, to check that if `localhost` resolves to `::1` ipv6, and if a candidate port is already allocated, `get_available_port` must skip that port

dhilloulinoracle (approved):

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #1594

dhilloulinoracle:
Merge Gate succeeded 🍏: Build ID #2343

PR #163, opened 2026-05-20
Improve tool decorator ergonomics

soufianeamini:
Using the tool decorator is not very ergonomic due to how the function signature is defined.
This PR adds overloads and tests to ensure that using it doesn't yield false positives while typechecking.

image

### DIFF
--- a/wayflowcore/src/wayflowcore/tools/toolhelpers.py
+++ b/wayflowcore/src/wayflowcore/tools/toolhelpers.py
@@ -455,7 +455,6 @@ def tool(
         and isinstance(func_or_name, str)
         and callable(func)
     ):
-        # if len(args) == 2 and (isinstance(args[0], str) and callable(args[1])):
         # Example case: wrapper with custom tool name
         # def my_callable():
         #     pass
@@ -471,7 +470,6 @@ def tool(
             requires_confirmation,
         )
     elif func_or_name is not None and isinstance(func_or_name, str):
-        # elif len(args) == 1 and isinstance(args[0], str):
         # Example case: decorator with custom tool name
         # @tool("my_callable1")
         # def my_callable():
--- a/wayflowcore/src/wayflowcore/tools/toolhelpers.py
+++ b/wayflowcore/src/wayflowcore/tools/toolhelpers.py
@@ -22,6 +22,7 @@ from typing import (
     Union,
     get_args,
     get_origin,
+    overload,
 )
 
 from wayflowcore.property import JsonSchemaParam, Property
@@ -79,7 +80,9 @@ def _get_partial_schema_from_annotation(arg_type: Type[Any]) -> JsonSchemaParam:
     # Handle Dict[str, X]
     if origin is dict or origin is Dict:
         if len(args) != 2:
-            raise TypeError("Dict must have exactly two type arguments, e.g., Dict[str, int]")
+            raise TypeError(
+                "Dict must have exactly two type arguments, e.g., Dict[str, int]"
+            )
         key_type, value_type = args
         if key_type is not str:
             raise TypeError("JSON object keys must be strings")
@@ -104,7 +107,10 @@ def _get_partial_schema_from_annotation(arg_type: Type[Any]) -> JsonSchemaParam:
         if len(unique_json_types) > 1:
             return {
                 "anyOf": [
-                    {"type": t, "enum": [v for v in args if PRIMITIVE_TYPE_MAP[type(v)] == t]}
+                    {
+                        "type": t,
+                        "enum": [v for v in args if PRIMITIVE_TYPE_MAP[type(v)] == t],
+                    }
                     for t in unique_json_types
                 ]
             }
@@ -141,7 +147,7 @@ def _get_tool_schema_no_parsing(
 
     if "self" in tool_signature.parameters.keys():
         raise TypeError(
-            f"The tool decorator cannot be used directly on a class method, use `tool(my_object.my_method)` instead"
+            "The tool decorator cannot be used directly on a class method, use `tool(my_object.my_method)` instead"
         )
 
     # Determining the schema of input parameters
@@ -205,7 +211,9 @@ def _get_tool_schema_from_parsed_signature(
                 f"Description mode is `infer_from_signature` but parameter {param_name} of tool {tool_name} is not Annotated "
                 f"(has type {annotated_param.annotation}). Either annotate the parameter or use the `only_docstring` description mode."
             )
-        param_annotation, param_description = _unpack_annotated_types(annotated_param.annotation)
+        param_annotation, param_description = _unpack_annotated_types(
+            annotated_param.annotation
+        )
         param_schema = _get_partial_schema_from_annotation(param_annotation)
 
         param_schema["description"] = param_description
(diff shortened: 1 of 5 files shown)


wayflowcore/tests/tools/test_tool_decorator.py
```diff
@@ -0,0 +1,31 @@
+from typing import Any, Callable
+from typing_extensions import assert_type
+from wayflowcore.tools.servertools import ServerTool
+from wayflowcore.tools.toolhelpers import DescriptionMode, tool
+
+
+def test_tool_decorator_result_type_is_correct() -> None:
+    tool_one = tool("tool_one")
```
dhilloulinoracle, line 18: hmmm, I am not sure to understand this use case.
What is supposed to happen in this case?
The tool is not invocable really is it?

soufianeamini: I'm preserving whatever semantics already existed, the general use-case is to use it to rename a tool. I added more comments to the tests to clarify things.

dhilloulinoracle: ok

dhilloulinoracle (approved):

wayflowcore/src/wayflowcore/tools/toolhelpers.py
```diff
@@ -231,8 +241,55 @@ def _get_tool_schema_from_parsed_signature(
...
+        DescriptionMode.INFER_FROM_SIGNATURE,
+        DescriptionMode.ONLY_DOCSTRING,
+        DescriptionMode.EXTRACT_FROM_DOCSTRING,
+    ] = DescriptionMode.INFER_FROM_SIGNATURE,
+    output_descriptors: Optional[List[Property]] = None,
+    requires_confirmation: bool = False,
+) -> Callable[[Callable[..., Any]], ServerTool]: ...
+
+
 def tool(
```
paul-cayet, line 280: This improves the named decorator/wrapper overloads, but it regresses the original case:

```python
@tool(description_mode=DescriptionMode.ONLY_DOCSTRING)
def simple_tool() -> None:
    """simple tool"""

```

Because func_or_name is now required, this raises before tool() can return the decorator partial. The same applies to documented forms like
@tool(description_mode="only_docstring"), @tool(output_descriptors=[...]), and @tool(requires_confirmation=True).

Can we restore the zero-positional decorator shape in both typing and runtime? It should be possible to:

- add a keyword-only overload returning `Callable[[Callable[..., Any]], ServerTool]`
- change the implementation signature to allow `func_or_name: Callable[..., Any] | str | None = None`
- add a `func_or_name is None` branch that returns the no-name partial decorator

One extra typing detail: the overload should accept both the `DescriptionMode.*` enum values and the documented string literal values (`"infer_from_signature"`, `"only_docstring"`, `"extract_from_docstring"`), otherwise the docs’ examples still produce type-checker false positives.

wayflowcore/tests/tools/test_tool_decorator.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # Wrapper with name and function passed as arguments
+    def func_two() -> None:
+        """Just a func"""
+
+    tool_two = tool("tool_two", func_two)
+    assert_type(tool_two, ServerTool)
+    assert isinstance(tool_two, ServerTool)
+
+    # Decorator with description mode
```
paul-cayet, line 49: This test uses `@tool("tool_three", description_mode=...)`, so it doesn’t cover the motivating call shape from the PR description, does it?

Can we add regression coverage for the keyword-only decorator form?

```python
@tool(description_mode=DescriptionMode.ONLY_DOCSTRING)
def simple_tool() -> None:
    """simple tool"""

```

Please assert both the runtime result and the type result, e.g. assert_type(simple_tool, ServerTool) and isinstance(simple_tool, ServerTool). I’d also include the documented string-literal form (description_mode="only_docstring") plus small cases for output_descriptors and requires_confirmation, since those all use the same zero-positional decorator path.

wayflowcore/src/wayflowcore/tools/toolhelpers.py
soufianeamini: Sorry for the oversight, I've restored the zero-positional decorator shape and made sure it worked at runtime and typing time. I also added the overload to the description mode with a DescriptionModeOptions type alias.

wayflowcore/tests/tools/test_tool_decorator.py
soufianeamini: I've added test cases for these options 👍

wayflowcore/tests/tools/test_tool_decorator.py
```diff
@@ -0,0 +1,114 @@
...
+    assert isinstance(another_func, ServerTool)
+    assert another_func.name == "real_function_name"
+
+    # Wrapper with name and function passed as arguments
+    def func_two() -> None:
+        """Just a func"""
+
+    tool_two = tool("tool_two", func_two)
+    assert_type(tool_two, ServerTool)
+    assert isinstance(tool_two, ServerTool)
```
sonleoracle, line 46: can you add an assert for the name too? same for the cases below if needed

wayflowcore/src/wayflowcore/tools/toolhelpers.py
```diff
@@ -390,47 +436,59 @@ def _make_tool(
...
+    if func is not None and isinstance(func_or_name, str) and callable(func):
         # Example case: wrapper with custom tool name
         # def my_callable():
         #     pass
         # my_tool = tool("my_callable1", my_callable)
-        # here args[0] is the tool name, and args[1] the callable
+        # here func_or_name is the tool name, and func the callable
         # we simply return the newly created ServerTool
-        tool_name = args[0]
+        tool_name = func_or_name
```
sonleoracle: do we need the `tool_name` variable at all?

wayflowcore/tests/tools/test_tool_decorator.py
```diff
@@ -0,0 +1,114 @@
...
+from typing_extensions import assert_type
+from wayflowcore.property import StringProperty
+from wayflowcore.tools.servertools import ServerTool
+from wayflowcore.tools.toolhelpers import DescriptionMode, tool
+
+
+def test_tool_decorator_result_type_is_correct() -> None:
+    # Wrapper with different name
+    tool_one = tool("tool_one")
+    assert_type(tool_one, Callable[[Callable[..., Any]], ServerTool])
```
sonleoracle, line 19: Please make sure this is actually tested and not suppressed since pyproject.toml mypy config ignores tests.*

sonleoracle (approved):

wayflowcore/tests/tools/test_tool_decorator.py
soufianeamini: I've added the assertions 👍

wayflowcore/src/wayflowcore/tools/toolhelpers.py
soufianeamini: Removed

wayflowcore/tests/tools/test_tool_decorator.py
soufianeamini: Done, I've special cased this file so that mypy typechecks it.

dhilloulinoracle:
Internal regression failed: Build ID #933

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #934

paul-cayet (approved):

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #1618

dhilloulinoracle:
Merge Gate failed: Build ID #2367

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #1619

dhilloulinoracle:
Merge Gate succeeded 🍏: Build ID #2368

PR #167, opened 2026-05-27
Handle text-encoded tool calls from OpenAI-compatible chat completions

paul-cayet:
## Description

Adds a Chat Completions fallback for OpenAI-compatible providers that return tool calls as assistant text content, such as LM Studio-style `<|tool_call>call:name{...}` markers, instead of structured `message.tool_calls` or `streamed delta.tool_calls`.

The fallback is applied only when structured tool calls are absent, safely parses JSON arguments, supports streaming and non-streaming responses, and preserves normal text output on malformed markers.


dhilloulinoracle:
original issue was fixed better by https://github.com/oracle/wayflow/pull/168

PR #169, opened 2026-05-28
improve Agent Spec tracing serialization and flow completion reporting

paul-cayet:
## Summary

Fixes WayFlow Agent Spec tracing compatibility issues by ensuring tracing spans/events are serialized with
the WayFlow Agent Spec plugin context, flow executions that end via `None` transitions still emit a valid
`branch_selected`, and shared tool-request transform serialization remains compatible with older configs.

## Changes

- Serialize Agent Spec tracing events/spans through `dump_tracing_model(...)` with WayFlow Agent Spec
plugins.
- Track the last executed flow step internally so flow-end tracing can populate `branch_selected` even
without a `CompleteStep`.
- Consolidate swarm/manager-workers tool request transforms into a shared implementation and register the
corresponding Agent Spec plugin transform.
- Add legacy upgrade paths for older Agent Spec and WayFlow-native serialized transform names.


dhilloulinoracle (approved):

dhilloulinoracle:
Internal regression failed: Build ID #907

wayflowcore/src/wayflowcore/agentspec/components/__init__.py
```diff
@@ -225,9 +225,9 @@
     "contextprovider_deserialization_plugin",
     "PluginAppendTrailingSystemMessageToUserMessageTransform",
     "PluginCoalesceSystemMessagesTransform",
+    "PluginToolRequestAndCallsTransform",
     "PluginRemoveEmptyNonUserMessageTransform",
     "PluginReactMergeToolRequestAndCallsTransform",
-    "PluginSwarmToolRequestAndCallsTransform",
```
sonleoracle, line 230: would this break existing user code that imports `PluginSwarmToolRequestAndCallsTransform`?

docs/wayflowcore/source/core/api/agentspec.rst
sonleoracle, line 235: should we add `PluginToolRequestAndCallsTransform` here?

wayflowcore/tests/serialization/test_swarm_serialization.py
```diff
@@ -245,3 +247,13 @@ def test_can_continue_a_deserialized_swarm_conversation(simple_swarm: Swarm) ->
     assert len(deser_conv.get_messages()) == conv_length_before_serialization
     deser_conv.append_user_message("Actually it's better now")
     deser_conv.execute()
+
+
+def test_legacy_swarm_transform_keeps_swarm_deserialization_behavior():
+    deserialized_template = deserialize(
+        type(_DEFAULT_SWARM_CHAT_TEMPLATE), serialize(_DEFAULT_SWARM_CHAT_TEMPLATE)
```
sonleoracle: the current `_DEFAULT_SWARM_CHAT_TEMPLATE` already contains the new transform, so this test is not testing legacy stuff. It should mutate the serialized config first, like the ManagerWorkers test does:

```python
serialized = serialize_to_dict(_DEFAULT_SWARM_CHAT_TEMPLATE)
serialized["post_rendering_transforms"][0]["_component_type"] = "_ToolRequestAndCallsTransform"

deserialized = deserialize(
    type(_DEFAULT_SWARM_CHAT_TEMPLATE),
    yaml.safe_dump(serialized, sort_keys=False),
)
```

wayflowcore/tests/agentspec/test_tracing.py
```diff
@@ -283,6 +290,131 @@ def test_agentspec_flow_raises_correct_events(wayflow_flow: Flow):
...
+    def say_hello(user_name: Annotated[str, "Name of the user"]) -> str:
+        """Return a greeting."""
+        return f"Hello {user_name}!"
+
+    manager_agent = Agent(name="manager", description="manager agent", tools=[say_hello], llm=llm)
+    worker = Agent(name="worker", description="worker", llm=llm)
+    group = ManagerWorkers(workers=[worker], group_manager=manager_agent)
+
+    listener = AgentSpecEventListener()
+    span_processor = DummyAgentSpecSpanProcessor()
```
sonleoracle: I think this span processor should serialize what tracing emits since this test is using plugin components:

```python
class SerializingSpanProcessor(DummyAgentSpecSpanProcessor):
      def on_event(self, event, span):
          super().on_event(event, span)
          dump_tracing_model(event)

      def on_end(self, span):
          super().on_end(span)
          dump_tracing_model(span)
```

wayflowcore/tests/agentspec/test_tracing.py
```diff
@@ -283,6 +290,131 @@ def test_agentspec_flow_raises_correct_events(wayflow_flow: Flow):
...
+
+    with patch_llm(
+        llm,
+        outputs=[
+            Message(
+                render_template(
+                    """
+{{thoughts}}
+
+{"name": {{tool_name}}, "parameters": {{tool_params}}}
```
sonleoracle, line 390: This template is not ideal:

- say_hello is not quoted, so it is not valid JSON.
- tool_params renders as a Python dict with single quotes, also not valid JSON.
- Because parsing fails, the message remains a normal agent/user-facing message instead of becoming a TOOL_REQUEST.

This is not caught because the assert is for `UserMessageRequestStatus` instead of tool request status or checking if the conversation contains tool request/tool result messages

sonleoracle (approved):

wayflowcore/src/wayflowcore/executors/executionstatus.py
```diff
@@ -66,7 +69,8 @@ def _deserialize_from_dict(
     ) -> "SerializableObject":
         return FinishedStatus(
             output_values=input_dict["output_values"],
-            complete_step_name=input_dict["complete_step_name"],
+            complete_step_name=input_dict.get("complete_step_name"),
+            _final_step_name=input_dict.get("_final_step_name", input_dict.get("final_step_name")),
```
sonleoracle: who would generate `final_step_name`? it does not seem wayflow is doing that anywhere, not even in main. it's only the test `test_finished_status_deserialization_supports_legacy_final_step_name` that is artificially creating this key:

```python
serialize(status).replace("_final_step_name", "final_step_name")
```

sonleoracle (approved):

wayflowcore/src/wayflowcore/agentspec/components/__init__.py
paul-cayet: Good point. I kept PluginSwarmToolRequestAndCallsTransform as an import-only alias to PluginToolRequestAndCallsTransform, while keeping the new canonical serialized name as PluginToolRequestAndCallsTransform.

docs/wayflowcore/source/core/api/agentspec.rst
paul-cayet: Added it to the Agent Spec API docs next to the other message transforms.

wayflowcore/tests/serialization/test_swarm_serialization.py
paul-cayet: Agreed. I updated the test to mutate the serialized config to _ToolRequestAndCallsTransform before deserialization, then assert it reserializes with ToolRequestAndCallsTransform.

wayflowcore/tests/agentspec/test_tracing.py
paul-cayet: Agreed. I added a SerializingSpanProcessor that calls dump_tracing_model(...) on emitted events/spans, so the test exercises the original plugin-component serialization failure mode.

wayflowcore/tests/agentspec/test_tracing.py
paul-cayet: Agreed. I changed the fixture to render valid JSON with json.dumps(...), preserved output parsing in the patched model path, and added assertions that real TOOL_REQUEST and TOOL_RESULT messages are produced.

vvenz1n (approved):

wayflowcore/src/wayflowcore/executors/executionstatus.py
paul-cayet: Good catch. final_step_name was not a real legacy key generated by WayFlow, only an artificial test artifact. I removed that fallback, deleted the artificial legacy test, and restored strict deserialization for complete_step_name.

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression failed: Build ID #1609

dhilloulinoracle:
Merge Gate failed: Build ID #2358

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #1610

dhilloulinoracle:
Merge Gate succeeded 🍏: Build ID #2359

PR #173, opened 2026-06-30
Feat: Exact Token Usage for OCI Streaming

sbackmann:
OCI Streaming did not collect usage metrics. This PR adds this.

### DIFF
--- a/wayflowcore/src/wayflowcore/models/ocigenaimodel.py
+++ b/wayflowcore/src/wayflowcore/models/ocigenaimodel.py
@@ -735,13 +735,13 @@ class _OciApiFormatter(ABC):
 
         cached_tokens = 0
         if "promptTokensDetails" in response and "cachedTokens" in (
-            response["promptTokensDetails"] or []
+            response["promptTokensDetails"] or {}
         ):
             cached_tokens = response["promptTokensDetails"]["cachedTokens"]
 
         reasoning_tokens = 0
         if "completionTokensDetails" in response and "reasoningTokens" in (
-            response["completionTokensDetails"] or []
+            response["completionTokensDetails"] or {}
         ):
             reasoning_tokens = response["completionTokensDetails"]["reasoningTokens"]
 
--- a/wayflowcore/src/wayflowcore/models/ocigenaimodel.py
+++ b/wayflowcore/src/wayflowcore/models/ocigenaimodel.py
@@ -526,6 +526,9 @@ class OCIGenAIModel(LlmModel):
         request = provider.convert_prompt_into_request(prompt, self.model_id)
         logger.debug(f"Streaming request to remote oci genai endpoint: {json.loads(str(request))}")
         request.is_stream = True
+        request.stream_options = oci.generative_ai_inference.models.StreamOptions(
+            is_include_usage=True
+        )
 
         return execute_sync_with_retry(
             lambda: self._client.chat(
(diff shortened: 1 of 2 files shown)


dhilloulinoracle (approved):

wayflowcore/src/wayflowcore/models/ocigenaimodel.py
```diff
@@ -719,6 +722,38 @@ def extract_usage(response: Dict[str, Any]) -> Optional[TokenUsage]:
...
         )
 
+    @staticmethod
+    def _extract_stream_usage(response_data: Dict[str, Any]) -> Optional[TokenUsage]:
+        response = response_data.get("usage", response_data)
+        if not isinstance(response, dict):
+            return None
+        if not all(
+            usage_key in response
+            for usage_key in ["promptTokens", "completionTokens", "totalTokens"]
```
sonleoracle, line 732: so if OCI does not return all 3, no token usage is logged at all? maybe it's better to do `if not any()` instead?

sonleoracle (approved):

dhilloulinoracle:
Internal regression failed: Build ID #945

wayflowcore/src/wayflowcore/models/ocigenaimodel.py
sbackmann: If OCI does not return all 3, it will return `None` (and thereby fallback to wayflow's approximate token count as before). That seems better to me than relying on a partially malformed OCI response -- but I can change it if you prefer the latter.

dhilloulinoracle:
Internal regression failed: Build ID #946

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #947

wayflowcore/src/wayflowcore/models/ocigenaimodel.py
sonleoracle: alright let's keep it as is

wayflowcore/src/wayflowcore/models/ocigenaimodel.py
```diff
@@ -719,6 +722,38 @@ def extract_usage(response: Dict[str, Any]) -> Optional[TokenUsage]:
...
+        ):
+            return None
+
+        cached_tokens = 0
+        if "promptTokensDetails" in response and "cachedTokens" in (
+            response["promptTokensDetails"] or []
```
paul-cayet: supernit: 

```suggestion
            response["promptTokensDetails"] or {}
```

and same below?

wayflowcore/src/wayflowcore/models/ocigenaimodel.py
```diff
@@ -1168,7 +1214,7 @@ def convert_oci_chunk_iterator_into_tagged_chunk_iterator(
                         accumulated_text,
                         text_content,
                     )
-                break
+                continue
```
paul-cayet, line 1217: why the change?

paul-cayet (approved):

wayflowcore/src/wayflowcore/models/ocigenaimodel.py
sbackmann: because the chunk with the token usage is emitted after the `finishReason` response, so we must not exit the loop here already.

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #1621

dhilloulinoracle:
Merge Gate succeeded 🍏: Build ID #2370

PR #175, opened 2026-07-13
Feat: Strict Structured Outputs

sbackmann:
Adds provider-independent strict structured-output validation.
- PromptTemplate(strict_output_validation=True) strictly validates prompt outputs; native structured generation fills omitted explicit defaults before validation.
- Agent(strict_output_validation=True) validates submit_result arguments, returns violations to the agent, and allows resubmission.
- Strict mode rejects wrong types, missing required fields, and unexpected fields without coercion.
- Explicit-default filling is separate from validation and shared across nested properties.
- Existing permissive normalization/coercion behavior is preserved.


wayflowcore/src/wayflowcore/agentspec/components/template.py
```diff
@@ -68,6 +68,8 @@ class PluginPromptTemplate(Component):
     """Specific format the llm answer should follow."""
     native_structured_generation: bool = True
     """Whether to use native structured generation or not. All llm providers might not support it."""
+    strict_output_validation: bool = False
+    """Whether Wayflow should reject, rather than coerce, invalid structured output."""
```
cesarebernardis: Nit: I would not mention wayflow in the docstring

wayflowcore/src/wayflowcore/executors/_agentexecutor.py
```diff
@@ -1102,6 +1104,12 @@ async def _execute_agent(
                 agent_state._get_current_tool_request()
             elif agent_state.curr_iter >= agent_config.max_iterations:
                 if len(agent_config.output_descriptors) > 0:
+                    if agent_config.agent_template.strict_output_validation:
```
cesarebernardis: I might need more context to understand why it's done here, but I think that validation should happen inside the model's generation, not here. The generation receives a Prompt object that should contain a flag saying whether strict output validation is required or not, and before returning an answer, validation happens and an exception is raised if it fails.

wayflowcore/src/wayflowcore/_outputvalidation.py
```diff
@@ -0,0 +1,122 @@
...
+        if output_name not in expected_by_name:
+            violations.append(f"{output_name}: unexpected field")
+
+    for output_name, output in expected_by_name.items():
+        if output_name not in outputs:
+            if output.has_default:
+                validated_outputs[output_name] = output.default_value
+            else:
+                violations.append(f"{output_name}: missing required field")
+            continue
```
cesarebernardis: Nit: maybe let's just use `else`

wayflowcore/src/wayflowcore/_outputvalidation.py
```diff
@@ -0,0 +1,122 @@
...
+    if violations:
+        raise StructuredOutputValidationError(violations=violations)
+    return validated_outputs
+
+
+def _validate_output(value: Any, output: Property, field_name: str) -> tuple[Any, list[str]]:
+    validated_value: Any
+    violations: list[str] = []
+
+    if isinstance(output, ObjectProperty):
```
cesarebernardis: I think it's better to make this logic part of the property classes, so that each property class is responsible for itself only, instead of having a long method

cesarebernardis: Note that properties already have `_is_value_of_expected_type`, but I guess it's not enough as it does not return the violations. You could avoid code duplication though, and make `_is_value_of_expected_type` rely on this once you move the logic.

wayflowcore/src/wayflowcore/exceptions.py
```diff
@@ -14,6 +14,14 @@ class WayFlowException(Exception):
     """Base exception for wayflowcore-related errors."""
 
 
+class StructuredOutputValidationError(WayFlowException, ValueError):
+    """Raised when locally validated structured LLM output is invalid."""
+
+    def __init__(self, violations: list[str]):
+        self.violations = violations
+        super().__init__("Structured output validation failed: " + "; ".join(violations))
```
cesarebernardis: Nit: might be easier to read with newlines

wayflowcore/src/wayflowcore/executors/_agentexecutor.py
sbackmann: The reason it is done here is because validating the agent output means validating the `submit_result` tool_args. I don't see how this can be moved into the model generation since at model-generation time, the agent may produce an ordinary message or any tool call. Also, note that the actual validation happens in `_collect_submit_tool_outputs` consistent with prior implementation, this here is just dealing with the case when the agent's `max_iterations` are exceeded.

wayflowcore/src/wayflowcore/_outputvalidation.py
sbackmann: I moved the recursive logic into the property classes. For duplication, primitive strict validation already uses `_is_value_of_expected_type`. `ObjectProperty` has different strict semantics though: regular validation allows extra keys and Python objects with matching attributes, whereas strict structured output does not. Since `ListProperty` and `DictProperty` can contain `ObjectProperty`, their strict traversal must also remain separate.

wayflowcore/src/wayflowcore/_outputvalidation.py
sonleoracle, line 1: Do we really need a new file for this?

wayflowcore/tests/integration/steps/test_prompt_execution_step.py
```diff
@@ -441,6 +442,20 @@ def test_structured_generation_returns_wrong_type(remotely_hosted_llm):
...
+    template = PromptTemplate.from_string(
+        template="Return the requested values.", strict_output_validation=True
+    )
+    step = PromptExecutionStep(
+        prompt_template=template,
+        llm=remotely_hosted_llm,
+        output_descriptors=[StringProperty(name="description")],
+    )
+    with patch_llm(remotely_hosted_llm, outputs=['{"description": ["first condition"]}']):
+        with pytest.raises(StructuredOutputValidationError, match="description"):
```
sonleoracle: Nit: the `match` message could be a bit longer

wayflowcore/tests/test_outputvalidation.py
```diff
@@ -0,0 +1,78 @@
...
+    descriptors = [
+        StringProperty(name="description"),
+        ObjectProperty(
+            name="details",
+            properties={"reason": StringProperty(name="reason", default_value="unknown")},
+        ),
+        StringProperty(name="optional", default_value="default"),
+    ]
+
+    assert validate_strict_outputs({"description": "condition", "details": {}}, descriptors) == {
```
sonleoracle: I would expect `validate_strict_outputs` to be an underscored function since it's private, and then I would not expect private functions to be tested directly, we would rather test through public APIs.

wayflowcore/src/wayflowcore/_outputvalidation.py
```diff
@@ -0,0 +1,46 @@
...
+            violations.append(f"{output_name}: unexpected field")
+
+    for output_name, output in expected_by_name.items():
+        if output_name not in outputs:
+            if output.has_default:
+                validated_outputs[output_name] = output.default_value
+            else:
+                violations.append(f"{output_name}: missing required field")
+        else:
+            value, value_violations = output._validate_strict_value(outputs[output_name])
```
sonleoracle: since each Property now has its own _validate_strict_value, do we need a separate `validate_strict_outputs` anymore?

wayflowcore/src/wayflowcore/executors/_agentexecutor.py
cesarebernardis: Yes I know, what I meant to say here is that this is a different scope compared to llm generation. The outputs of an agent are built differently, they are not dependent on the outputs you define in the prompt template. Feels weird that the `strict_output_validation` you pass to the agent's template applies to the outputs of the agent, while the outputs in it are ignored. Maybe we should have the same flag at the agent's init level, if you need this strict generation also out of the agent

sbackmann: But wouldn't the same reasoning apply to the `PromptExecutionStep` as well? There I can also pass `strict_output_validation` via the `PromptTemplate` and define the outputs via the step's `output_descriptors` right?

That said, I'm fine with passing it as an init level argument instead. I'm just wondering whether in that case it should be removed from the `PromptTemplate` altogether and be supplied as an init level argument in both cases.

wayflowcore/src/wayflowcore/_outputvalidation.py
cesarebernardis: > ObjectProperty has different strict semantics though: regular validation allows extra keys

Sounds problematic to me, I would expect an ObjectProperty to not have extra keys, considering the meaning that it has in waylfow. It could be considered a bug.

> and Python objects with matching attributes

Why this is not accepted in strict? It's arguably not possible to have one, but it would match the type.

wayflowcore/src/wayflowcore/executors/_agentexecutor.py
cesarebernardis: The `PromptExecutionStep` is a bit different, because the whole point of the step is to run the given prompt through an LLM. Indeed, it takes the prompt's object response format to build output descriptors, and it raises an error if you have both explicit output descriptors and a response format in the prompt and they don't match exactly.

I think it's correct to have it at Prompt level, and that should be enforced at the level where the prompt is used, i.e., in the model's `generation` call. Then there are other layers where type is enforced, like the agent, where structured output might be built over multiple calls, so in that case we probably need an additional flag that enforces correct typing in a different way.

Do you need it at the agent level for your use case?

wayflowcore/src/wayflowcore/_outputvalidation.py
sbackmann: > Why this is not accepted in strict? It's arguably not possible to have one, but it would match the type.
Yes, the rationale was it's not possible to have an LLM response with a python object. But I can add it.

However, it would be difficult in handling the explicit defaults for python objects with partially missing fields. We would have to either:
- mutate or create a new python object during validation where we apply the defaults -- I don't think we can generally, safely do that.
- reject python objects with missing fields even if they have a default value supplied.

> I would expect an ObjectProperty to not have extra keys, considering the meaning that it has in waylfow. It could be considered a bug.

It seems intentional to me given that it is explicitly set for the openai case [here](https://github.com/oracle/wayflow/blob/main/wayflowcore/src/wayflowcore/property.py#L962). I can change that but it would be a breaking change.

wayflowcore/src/wayflowcore/executors/_agentexecutor.py
sbackmann: Yes, the agent is the main use case.

wayflowcore/src/wayflowcore/_outputvalidation.py
sbackmann: Do you have a concrete suggestion? For now, I have moved it into `property.py`.

wayflowcore/tests/test_outputvalidation.py
sbackmann: with the move into `property.py` it's now a public method in a public module, so let me know if this concern is resolved.

wayflowcore/src/wayflowcore/_outputvalidation.py
sbackmann: The only way I see to get rid of it would be to create a wrapping `ObjectProperty` and call `_validate_strict_value` on that which to me seems like an uglier hack than having one additional helper method. Let me know what you think.

wayflowcore/tests/test_outputvalidation.py
sonleoracle: but I still see in the code that `validate_strict_outputs` is defined in `_outputvalidation.py` which seems to be a private module

wayflowcore/src/wayflowcore/_outputvalidation.py
sonleoracle: Yes moving it there seems to be a better place, and please add an underscore to the function name

wayflowcore/src/wayflowcore/_outputvalidation.py
sonleoracle: ok I guess we can keep it as is

sonleoracle (approved):

wayflowcore/src/wayflowcore/_outputvalidation.py
sbackmann: Really, but this is used across different classes?

wayflowcore/tests/test_outputvalidation.py
sbackmann: In the change tab of this PR there is no `_outputvalidation.py` anymore.

wayflowcore/src/wayflowcore/_outputvalidation.py
sonleoracle: In wayflowcore, if some API is not meant to be used by end-users, it is a private API, and should be prefixed with an underscore

sbackmann: ok, changed.

dhilloulinoracle:
Internal regression failed: Build ID #965

wayflowcore/src/wayflowcore/property.py
```diff
@@ -115,6 +116,19 @@ def is_value_of_expected_type(self, value: Any) -> bool:
             return False
         return self._is_value_of_expected_type(value)
 
+    def _validate_strict_value(self, value: Any) -> tuple[Any, list[tuple[str, str]]]:
```
cesarebernardis: `Validate` functions probably should not return the value, it's always the same one given as input

wayflowcore/src/wayflowcore/property.py
```diff
@@ -115,6 +116,19 @@ def is_value_of_expected_type(self, value: Any) -> bool:
...
+        """Validate a value without coercion or implicit defaults."""
+        if self.is_value_of_expected_type(value):
+            return value, []
+        return value, [("", self._strict_type_violation(value))]
+
+    def _fill_explicit_defaults(self, value: Any) -> Any:
```
cesarebernardis: Can we call it something like `_fill_nested_values_with_explicit_defaults`? 

I struggled understanding what this method was doing, and it is skipping 1 level of default filling, as properties are not looking at themselves to see if they have a default, they look at inner properties (if they have any).

wayflowcore/src/wayflowcore/property.py
```diff
@@ -847,6 +880,29 @@ def _is_value_of_expected_type(self, value: Any) -> bool:
...
+            )
+            violations.extend(
+                (f".{key}{location}", message) for location, message in nested_violations
+            )
+        return validated_value, violations
+
+    def _fill_explicit_defaults(self, value: Any) -> Any:
+        if not isinstance(value, dict):
+            return value
+        return {key: self.value_type._fill_explicit_defaults(item) for key, item in value.items()}
```
cesarebernardis: Might be absurd and unlikely to happen, but also the key is a potentially complex property

wayflowcore/src/wayflowcore/models/llmmodel.py
```diff
@@ -67,6 +68,17 @@ class Prompt(SerializableDataclassMixin, SerializableObject):
...
 
     def parse_output(self, message: "Message") -> "Message":
+        if self.response_format is not None:
+            try:
+                value = json.loads(message.content)
+            except json.JSONDecodeError:
+                pass
+            else:
+                filled_value = self.response_format._fill_explicit_defaults(value)
+                if filled_value != value:
```
cesarebernardis: I think we can avoid checking equality, as it might be complicated. You can just replace with `filled_value` at this point

cesarebernardis (approved):

wayflowcore/src/wayflowcore/property.py
sbackmann: True, but I'm not sure we should handle this here. None of the built-in complex properties can be a valid dictionary key since they're unhashable. And for any other potential future properties we would have to guarantee that the keys are still hashable and unique after our default filling.

dhilloulinoracle (approved):

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #968

dhilloulinoracle:
Merge Gate failed: Build ID #2378

dhilloulinoracle:
Merge Gate failed: Build ID #2379

wayflowcore/src/wayflowcore/property.py
sbackmann: Discussed offline

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #1627

dhilloulinoracle:
Merge Gate succeeded 🍏: Build ID #2380

PR #178, opened 2026-07-15
additionalProperties in ObjectProperty

cesarebernardis:
JSON Schema additionalProperties support in ObjectProperty.

JSON schema has an attribute in the `object` type that is called `additionalProperties`. It defines whether properties that are not described in the JSON schema’s `properties` attribute are accepted (and optionally what types are accepted) or not. So far wayflow’s `ObjectProperty` was assuming `True` for this in the validation methods it implements. Some use cases require tighter control on this.

### DIFF
--- a/wayflowcore/src/wayflowcore/property.py
+++ b/wayflowcore/src/wayflowcore/property.py
@@ -1019,12 +1019,14 @@ class ObjectProperty(Property):
             # https://platform.openai.com/docs/guides/structured-outputs#all-fields-must-be-required
             if self.additional_properties:
                 json_schema["additionalProperties"] = False
-                warnings.warn(
-                    "additionalProperties value different from False is not compatible "
-                    f"with OpenAI APIs; The value will be ignored.",
-                    UserWarning,
-                    stacklevel=2,
-                )
+                if isinstance(self.additional_properties, Property):
+                    # We don't warn if True because it's the default value
+                    warnings.warn(
+                        "additionalProperties are not compatible with OpenAI APIs; "
+                        f"The value will be ignored.",
+                        UserWarning,
+                        stacklevel=2,
+                    )
             json_schema["required"] = list(json_schema["properties"].keys())
             for property_name, property_ in json_schema["properties"].items():
                 if "default" in property_:
--- a/wayflowcore/src/wayflowcore/property.py
+++ b/wayflowcore/src/wayflowcore/property.py
@@ -914,7 +914,7 @@ class ObjectProperty(Property):
         Dictionary of property names and their types. Defaults without any property.
     additional_properties:
         Boolean indicating if the object accepts arbitrary properties, or Property that defines which types
-        of additional properties are accepted.
+        of additional properties are accepted. Default is True.
     """
 
     properties: Dict[str, "Property"] = field(default_factory=dict)
--- a/wayflowcore/src/wayflowcore/property.py
+++ b/wayflowcore/src/wayflowcore/property.py
@@ -342,6 +342,8 @@ class Property(SerializableObject, ABC):
             return AnyProperty(**kwargs)
 
         if "properties" in schema:
+            # JSON Schema allows an object to define named fields and a rule for other fields.
+            additional_properties = schema.get("additionalProperties", True)
             return ObjectProperty(
                 **kwargs,
                 properties={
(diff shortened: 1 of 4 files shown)


dhilloulinoracle:
Internal regression failed: Build ID #948

wayflowcore/src/wayflowcore/property.py
```diff
@@ -902,19 +912,47 @@ class ObjectProperty(Property):
...
         )
+        if not known_properties_are_valid:
+            return False
+
+        # Dictionaries have keys, while Python objects keep their instance attributes in __dict__.
+        # Values without either form have no additional properties to check.
+        all_values = (
+            value.items()
+            if isinstance(value, dict)
+            else vars(value).items() if hasattr(value, "__dict__") else ()
```
sbackmann, line 940: maybe niche case but this would miss python objects that have `__slots__` defined and thus could have undeclared attributes but no `__dict__`.

wayflowcore/src/wayflowcore/property.py
```diff
@@ -902,19 +912,47 @@ class ObjectProperty(Property):
               the flow can always execute properly with some defaults if needed.
     properties:
         Dictionary of property names and their types. Defaults without any property.
+    additional_properties:
+        Boolean indicating if the object accepts arbitrary properties, or Property that defines which types
+        of additional properties are accepted.
     """
 
     properties: Dict[str, "Property"] = field(default_factory=dict)
+    additional_properties: Union[bool, Property] = True
```
sbackmann, line 921: Wouldn’t this break serialization [here](https://github.com/oracle/wayflow/blob/main/wayflowcore/src/wayflowcore/serialization/_builtins_serialization_plugin.py#L616)? It assumes `additionalProperties` is a dictionary but it can now be a boolean.

sbackmann (approved):

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #950

wayflowcore/src/wayflowcore/property.py
cesarebernardis: Unrelated to this PR, since that method works with pyagentspec's `Property` json schemas, not wayflow's ones, but that serialization part is imprecise anyway. I'll update it.

wayflowcore/src/wayflowcore/property.py
cesarebernardis: That's a pretty niche scenario, I would keep the current implementation for simplicity

wayflowcore/src/wayflowcore/property.py
```diff
@@ -902,19 +912,47 @@ class ObjectProperty(Property):
               the flow can always execute properly with some defaults if needed.
     properties:
         Dictionary of property names and their types. Defaults without any property.
+    additional_properties:
+        Boolean indicating if the object accepts arbitrary properties, or Property that defines which types
```
dhilloulinoracle, line 916: please indicate what it defaults to

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #951

dhilloulinoracle (approved):

wayflowcore/src/wayflowcore/property.py
```diff
@@ -902,19 +912,47 @@ class ObjectProperty(Property):
...
+            else vars(value).items() if hasattr(value, "__dict__") else ()
+        )
+        # Named properties were checked above. Only validate the remaining values here.
+        additional_values = (
+            property_value
+            for property_name, property_value in all_values
+            if property_name not in self.properties
+        )
+        if self.additional_properties is False:
+            return not any(True for _ in additional_values)
```
sonleoracle, line 949: It's not immediately clear this `any(True for...)` is because `additional_values` was declared as a generator. Should `additional_values` just become a regular list, with the square brackets?

sonleoracle (approved):

wayflowcore/src/wayflowcore/property.py
cesarebernardis: This is to lazily materialize the values only if they are actually required, as we care only if there's at least one additional value, not which one(s) or how many in total.

wayflowcore/src/wayflowcore/property.py
cesarebernardis: Done

dhilloulinoracle:
Merge Gate INPROGRESS

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #1620

dhilloulinoracle:
Merge Gate succeeded 🍏: Build ID #2369

Issue #155, opened 2026-05-06
StartNode DFE variables not bound when InputMessageNode is first gate — downstream nodes receive empty values


groganz:
## Summary

When a `WayflowFlow` has an `InputMessageNode` as its first gate (the bug described in #151), any data flow edges (DFEs) originating from `StartNode` are never populated. Downstream nodes — such as `LlmNode` or `ApiCallNode` — receive empty strings or `None` for all StartNode-sourced variables, even when the triggering `message/send` call carries those values in the task input.

## Root cause

The pre-execute bug (#151) means `StartNode` is never processed against the trigger payload: the pre-execute runs the flow before the incoming message (which carries the StartNode inputs) is appended, so `StartNode.output_data` is empty when `execute_async()` first runs. Any DFE edges wired from `StartNode → LlmNode.system_message` or `StartNode → ApiCallNode.body` therefore receive empty values for the entire lifetime of the task — including after the user later resumes the task via subsequent `message/send` calls.

## Steps to reproduce

1. Create a `WayflowFlow` where:
   - `StartNode` has one or more `input_descriptors` (e.g. `run_id: str`).
   - The first node after `StartNode` is an `InputMessageNode`.
   - A later `LlmNode` or `ApiCallNode` has a DFE wired from `StartNode.run_id`.
2. Send an initial `message/send` with `{"input": {"run_id": "abc123"}}`.
3. Resume the task after the `InputMessageNode` gate.
4. Observe: the `LlmNode`/`ApiCallNode` receives `""` or `None` for `run_id` instead of `"abc123"`.

## Expected behaviour

Values supplied in the initial `message/send` input should be bound to `StartNode.output_data` and propagated along DFE edges regardless of whether an `InputMessageNode` appears as the first gate.

## Actual behaviour

`StartNode.output_data` is empty because the node processes its inputs during the pre-execute phase — before the trigger message is appended to the conversation. All DFE-sourced variables from `StartNode` arrive at downstream nodes as empty strings for the full duration of the task.

## Impact

Any integration that relies on passing per-task metadata (e.g. correlation IDs, tenant IDs, run identifiers) through `StartNode` DFE edges into `LlmNode` or `ApiCallNode` is broken when an `InputMessageNode` is the first gate. The workaround is to inject the metadata through some other mechanism (e.g. system-message templating outside the flow, or a custom `ContextVar` propagation layer), which defeats the purpose of DFE wiring for these values.

## Context

- `wayflowcore` 26.1.1
[...]

PR #187, opened 2026-08-26
fix: don't pre-execute a flow past its first InputMessageStep

### LINKED ISSUE #151: A2AAgentWorker pre-executes flow before user's first message, causing InputMessageStep to consume trigger message as form data
## Summary

When a `WayflowFlow` containing an `InputMessageStep` receives its first A2A `message/send` call, `A2AAgentWorker.run_task` pre-executes the conversation before appending the user's message:

```python
if isinstance(self.assistant, WayflowFlow) and any(
    isinstance(step, InputMessageStep) for step in self.assistant.steps.values()
):
    await conversation.execute_async()
```

The pre-execute runs the flow until it yields at the first `InputMessageStep` (setting `asked_user = True` in the step's internal context). The worker then **immediately** appends the user's incoming message and calls `execute_async()` again. Because `asked_user` is now `True`, the `InputMessageStep` treats that incoming message as the user's form response rather than a trigger — and passes it downstream without ever returning `input-required` to the caller.

## Steps to reproduce

1. Build a `WayflowFlow` whose first node (after `StartNode`) is an `InputMessageStep` that expects structured input (e.g. required fields in its schema).
2. Send any initial `message/send` request — a trigger message that does **not** contain the expected form fields.
3. Observe: the `InputMessageStep` consumes the trigger message as if it were the user's form submission and passes its raw content to the next step, instead of returning `state: input-required` so the client can collect proper input from the user.

## Expected behaviour

When the flow reaches an `InputMessageStep` and the user has not yet provided the expected input, the A2A task should transition to `input-required`. The client gets a chance to render the appropriate UI, collect the user's response, and resume the task with the actual form data.

## Actual behaviour

The pre-execute sets `asked_user = True`, so the very next `execute_async()` call (triggered by the trigger message) skips the "ask" phase and immediately treats the trigger message as the user's answer. If the trigger message does not contain the expected fields, downstream steps receive incomplete or wrong data, typically resulting in a runtime error and `state: failed`.

## Context

- `wayflowcore` 26.1.1  
- `A2AAgentWorker.run_task` in `wayflowcore/agentserver/a2a/_worker.py`  
- The pre-execute pattern makes sense when there are **non-interactive** steps (e.g. `ApiNode`, `LlmNode`) before the first `InputMessageStep`. In that case, running up to the first yield is useful. But when the `InputMessageStep` is the **first** meaningful step in the flow, the pre-execute is a no-op that leaves the step in the wrong state, causing the trigger message to be misinterpreted.

## Suggested fix

Guard the pre-execute on whether any non-interactive steps actually precede the first `InputMessageStep`. If the first `InputMessageStep` is reached immediately (no intermediate work), skip the pre-execute so the first user message hits the step in its initial state and the flow can correctly return `input-required`.

Alternatively, document clearly that callers are expected to supply the `InputMessageStep`'s required input in the **first** `message/send` call, and that returning `input-required` on the first call is not supported for `WayflowFlow` agents.

mittalpk:
Fixes #151.

`A2AAgentWorker.run_task` pre-executes a brand-new conversation whenever the assigned flow contains an `InputMessageStep` anywhere, so that any non-interactive steps get a chance to run before the first yield. The problem: pre-execution always runs before the caller's actual first message is ever appended to the conversation. If that pre-execution reaches and yields at the `InputMessageStep` itself, the worker immediately appends the caller's real first message and re-executes -- and the `InputMessageStep` (which tracks a simple "have I already asked" flag) treats that message as the answer to a question the caller was never actually shown, instead of returning `input-required` and giving the caller a real chance to respond.

This reproduces with the repo's own existing `FLOW_THAT_WITH_INPUT_STEP_YIELDS_ONCE` test fixture -- `test_server_returns_error_when_sending_completed_task` currently passes with the very first `message/send` call ("John") landing directly in `state: completed`, with no `input-required` round-trip ever happening.

The fix adds `_flow_yields_at_input_step_before_any_real_message()`, which walks the flow's single unambiguous execution path from its begin step. If that path reaches an `InputMessageStep` (with or without other, non-yielding steps running first) before branching or hitting a dead end, pre-execution is skipped so the caller's first message reaches the `InputMessageStep` in its real initial state. If the path branches or can't be resolved unambiguously, the existing pre-execute behavior is conservatively kept.

I also updated `test_server_returns_error_when_sending_completed_task`, since its first assertion encoded the buggy behavior (`state == "completed"` after a single message) -- it now expects `input-required` after the first message, then completes the task with a follow-up message before continuing with the rest of the test.

Note on scope: issue #155 (StartNode DFE variables not bound when InputMessageNode is first gate) is filed as the same root cause as #151, but on closer inspection it isn't -- `A2AAgentWorker.run_task` never passes `inputs=` to `start_conversation()` at all, regardless of pre-execute timing, so there's no existing convention for conveying typed `StartNode` inputs over the A2A protocol to build on. That looks like a separate feature request needing a maintainer decision on the wire format, not a fix to this bug. I haven't touched it in this PR; happy to help once there's a direction agreed for it.

Verification:
- New unit tests (`tests/a2a/test_worker.py`) directly cover `_flow_yields_at_input_step_before_any_real_message()`: immediate InputMessageStep, a non-yielding step before it, no InputMessageStep at all, and an ambiguous branching case.
- Updated `test_server_returns_error_when_sending_completed_task` passes with the fix, confirmed to fail against unpatched code (`state == "input-required"` vs the old `"completed"`).
- Full `tests/a2a/` suite: no new failures introduced. All Flow/deterministic tests pass; the pre-existing failures that need a real LLM backend (`AGENT_WITH_SERVER_TOOL`, `test_a2aagent.py`, etc.) are identical on a clean `main` checkout with the same test environment, confirmed via direct comparison.
- `black`, `isort`, `mypy` (repo's pinned versions, root `pyproject.toml` config) all clean on the changed files.

### DIFF
--- a/wayflowcore/src/wayflowcore/agentserver/a2a/_worker.py
+++ b/wayflowcore/src/wayflowcore/agentserver/a2a/_worker.py
@@ -31,6 +31,39 @@ Context = Any
 """The shape of the context stored in the storage (is not used in our case but is added to be compatible with fasta2a's Worker class"""
 
 
+def _flow_yields_at_input_step_before_any_real_message(flow: WayflowFlow) -> bool:
+    """Whether pre-executing ``flow`` (before any real message has been
+    appended) would run all the way to an ``InputMessageStep`` and yield
+    there.
+
+    Pre-execution always runs on a brand-new conversation, before the
+    caller's first message is ever appended. If that pre-execution reaches
+    and yields at an ``InputMessageStep`` — whether or not other,
+    non-yielding steps ran first — the worker then immediately appends the
+    caller's first message and re-executes, which the ``InputMessageStep``
+    misinterprets as the answer to a question the caller was never actually
+    asked. We only walk a single, unambiguous path (a step with more than one
+    distinct outgoing destination is treated as "can't tell", and we
+    conservatively keep the existing pre-execute behaviour).
+    """
+    visited: set[int] = set()
+    step = flow.begin_step
+    while id(step) not in visited:
+        visited.add(id(step))
+        if isinstance(step, InputMessageStep):
+            return True
+        destinations = {
+            edge.destination_step for edge in flow.control_flow_edges if edge.source_step is step
+        }
+        if len(destinations) != 1:
+            return False
+        (destination,) = destinations
+        if destination is None:
+            return False
+        step = destination
+    return False
+
+
 @dataclass
 class A2AAgentWorker(Worker[Context]):  # type: ignore[misc]
     """A worker that uses a WayFlow conversational component to execute tasks"""
@@ -77,8 +110,12 @@ class A2AAgentWorker(Worker[Context]):  # type: ignore[misc]
             messages = None if conversation is None else conversation.message_list
             conversation = self.assistant.start_conversation(messages=messages)
 
-            if isinstance(self.assistant, WayflowFlow) and any(
-                isinstance(step, InputMessageStep) for step in self.assistant.steps.values()
+            if (
+                isinstance(self.assistant, WayflowFlow)
+                and any(
+                    isinstance(step, InputMessageStep) for step in self.assistant.steps.values()
+                )
+                and not _flow_yields_at_input_step_before_any_real_message(self.assistant)
             ):
                 await conversation.execute_async()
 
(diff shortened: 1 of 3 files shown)


cesarebernardis:
@mittalpk Thank you for the contribution!

To start, could you please sign the Oracle Contributor Agreement (OCA)?

You can find more info in the comment at https://github.com/oracle/wayflow/pull/187#issuecomment-5431588444. Please make sure to use the same email associated to your Github account, that should be the one reported in the comment.

mittalpk:
Hello @cesarebernardis, i have signed Oracle Contributor Agreement (OCA) on 26.08.2026. it is still under review.

PR #193, opened 2026-09-11
Reject dynamic dicts in OpenAI strict schemas

sonleoracle:
fix structured output schema to be compatible to latest OpenAI models

### DIFF
--- a/wayflowcore/src/wayflowcore/property.py
+++ b/wayflowcore/src/wayflowcore/property.py
@@ -908,6 +908,14 @@ class DictProperty(Property):
         return dict()
 
     def _type_to_json_schema(self, openai_compatible: bool = False) -> JsonSchemaParam:
+        if openai_compatible:
+            # OpenAI strict structured outputs require every object to set
+            # additionalProperties to false, which is incompatible with an arbitrary map.
+            raise ValueError(
+                "DictProperty is not supported by OpenAI strict structured output. "
+                "Use an ObjectProperty with fixed fields or a ListProperty of key/value objects instead."
+            )
+
         # does not support non string keys
         if not isinstance(self.key_type, (StringProperty, AnyProperty)):
             raise ValueError(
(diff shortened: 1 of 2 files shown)


dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #1028

cesarebernardis (approved):

wayflowcore/src/wayflowcore/property.py
```diff
@@ -908,6 +908,14 @@ def _type_default_value(self) -> Any:
         return dict()
 
     def _type_to_json_schema(self, openai_compatible: bool = False) -> JsonSchemaParam:
+        if openai_compatible:
```
cesarebernardis, line 911: Arguably openai models don't support them, other `compatible` deployments might accept them, am I wrong? Don't we have another more precise method for openai (only) models?

cesarebernardis (approved):

wayflowcore/src/wayflowcore/property.py
dhilloulinoracle: yeah might want to use an argument with a different name and check what was the intended meaning of the "openai_compatible" argument before

sonleoracle: Originally, `openai_compatible` was meant for supporting compatibility with OpenAI's structured generation in strict mode, to disallow additionalProperties and nullable fields in the ObjectProperty, but this argument's actual usage also applies to other OpenAI-compatible models such as `VllmModel` (which does not need this restriction).

But yes, raising the DictProperty for all OpenAI-compatible models was too strict. I added a new parameter `openai_strict` just for rejecting this DictProperty for OpenAI endpoints.

paul-cayet (approved):

dhilloulinoracle:
Internal regression failed: Build ID #1044

dhilloulinoracle:
Internal regression succeeded 🍏: Build ID #1045

